PR #36, opened 2017-02-01
Did some changes to make it cross platform compatible 

ThePerry-2015:
Did some changes to make it cross platform compatible regarding icons and packaging instructions.
Forgot that I hadn't updated my fork in a while so it attached a ton of updates and stuff in the same run, but it was rebased on your last update to the development before I did my changes so it should be the same plus my last changes in package.json and main.js.


ThePerry-2015:
Yeah, I have to be better at keeping the fork updated for each update you do so that it only shows the ones I've actually been adding and not everything you have already added into the code between the versions.

PR #71, opened 2017-02-20
added firebotlog and fireboterror

TheBaconSpace:


Firebottle:
Thanks Bacon! I think this might make it a bit confusing though as console.log is pretty common place and putting that into a const would confuse people as to the purpose of "firebotlog".

PR #72, opened 2017-02-21
Allow Auth with Microsoft, Twitter, & Discord

### LINKED ISSUE #63: Alt Login Improvements
Last release I added in an alternative login path that took you to a page to copy/paste your token into the app. This fixed login issues with microsoft, twitter, and discord. However, Deeky let me know there is a fix to make these services work with the regular login path. I don't want to remove the alternative just in case, but this should be added into the next release too.

**Deeky Notes**
if you were receiving the issue where the popups for Microsft, Twitter, and DIscord were empty white screens, all you need to do is modify your windowParams to have webPreferences: { sandbox: true }

DeekyJay:
Just thought I'd quickly fire this here to get it out of the way!
Resolves #63

### DIFF
--- a/lib/login/login.js
+++ b/lib/login/login.js
@@ -23,7 +23,8 @@ ipcMain.on('oauth-login', function(event, streamOrBot) {
         show: true,
         webPreferences: {
             nodeIntegration: false,
-            partition: 'persist:'+date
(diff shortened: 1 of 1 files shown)


Firebottle:
Oh! Thanks so much! You're the best. :)

PR #142, opened 2017-06-03
Switch Firebot to electron-oauth2 package 

kateract:
Sorry for the spacing changes, I use ALT-SHIFT-F sometimes.


Firebottle:
Thanks! Will take a look tomorrow. One good thing about moving to this is that we can add in the bypass slow chat and bypass links scopes again. That will fix problems with the API buttons and chat buttons being chained.

Issue #52, opened 2017-02-07
Button: Custom Scripts / Custom Web API calls


Firebottle:
Add in a new button that, when clicked, will load up a JS file and run a function. This would allow people to run custom scripts. This would be a very dangerous option for most people, so we'd need to have people specifically enable it in the settings area with some type of warning.

ebiggz:
I am currently working on this feature. Just commenting here to make sure we don't have multiple folks working on this in tandem. I am nearing completion and will have a pull request in soon. 👍

Firebottle:
@ebiggz holy crap this is super cool. Testing it all now and will hopefully get it into the next build.

ebiggz:
@Firebottle Awesome, that example is great.

Firebottle:
Released 3.3.0

PR #169, opened 2017-06-07
Custom Script Button Effect

ebiggz:
This adds the Custom Script button effect.

This effect will call a function on the selected JS script and will pass the **buttonId** of the button that was clicked and the **username** of the viewer that clicked it. The custom script can optionally return a `Promise` that contains a string message to chat as either the streamer or the bot and can even designate the chat message to instead be a whisper to the person who clicked the button. 

Upon app startup, a "scripts" folder will be created in the root directory of Firebot, which will be searched through for any .js files when generating a list of available scripts. I opted for this method instead of a file picker as I felt it forces the user to keep their scripts centralized and allows savvy script writers to always know the script's relative location. That could be particularly useful if they are writing a script for someone else to use and don't have to guess where the script might be located. However, if we think a regular file picker is a better option, it will be super easy to change.

image

This also adds a new setting to the Settings tab, so the user is required to opt-in to use this feature (as suggested in Firebottle's original request #52) . I think the opt-in requirement, the warning in the description of the opt-in, and the final warning when adding a custom script effect to a button should be an ample amount of warnings.

image

~~We will need to create a wiki section on how to make a valid custom script.~~ (Edit: a wiki page has been [created](https://github.com/Firebottle/Firebot/wiki/Writing-Custom-Scripts).) But I'll briefly document what is required at a bare-bones level here:

```javascript
function run(buttonId, username) {
    // Do some stuff
}

// Export 'run' function so it is visible to Node
exports.run = run;
```

And here is a more advanced example: 
```javascript
function run(buttonId, username) {
  // Return a Promise object
  return new Promise((resolve, reject) => {
    // Do some stuff

    // Create a response 
    var response = {
      success: true,
      message: "Yay, we ran a custom script!",
      chatAs: "streamer",
      whisper: false
    }

    // Resolve Promise with the response object
    resolve(response);
  });
}

// Export 'run' function so it is visible to Node
exports.run = run;
```
~~Once we create a wiki entry, it wouldn't hurt to add a link to it in the opt-in setting description text or something.~~ (Edit: this has been [added](https://github.com/Firebottle/Firebot/pull/169/commits/45df5bdcf0e1d76849214104d94a46bedef4eda3), too)

As far as testing goes, I've verified the new GUI elements and script execution works. The only thing I haven't tested yet is messages from scripts getting relayed to chat or as a whisper. I couldn't test because I was having unrelated issues getting my bot to connect to my channel. I am fairly confident everything should work as I am just reusing the `mixer-chat` module but it wouldn't hurt to double check anyway.

We may also want to spot check the rest of the app as I had to do a manual merge because there was some conflicts. So far everything looks good to me, but a second pair of eyes would be great.

And that's about it.
Hit me with your questions and suggestions if you got them!


lib/interactive/control-router.js
```diff
@@ -187,4 +209,4 @@ ipcMain.on('manualButton', function(event, controlID) {
 });
 
 // Export Functions
-exports.router = controlRouter;
\ No newline at end of file
+exports.router = controlRouter;
```
ebiggz: I'm not sure why random whitespace chacters are still getting stripped. I thought I turned that off in my IDE. Sorry about this!

Issue #167, opened 2017-06-06
Cooldown Tracking


Firebottle:
The app currently does not keep track of any cooldown state and just sends whatever is provided. So, if you press a button that puts it on cooldown for 3 minutes, and then another that puts that same button on cool down for 10 seconds, the button will flip from 3 minutes to the 10 second cooldown.

To fix this we'd need to keep track of cooldowns for all buttons and then only send a cooldown if it is greater than what the button is at currently.

Firebottle:
https://github.com/Firebottle/Firebot/releases/tag/v3.4.0

PR #176, opened 2017-06-08
Cooldown saving.

Firebottle:
Addresses #167 

Cooldowns are now only applied if the new cooldown amount would make people wait longer than the current cooldown.

For example, if someone hits a button that cools something down for 3 minutes. You can no longer hit another button that would change that to, say, 10 seconds. This helps when you have multiple different ways to cool down a button. The app will pick whatever wait is longest.

### DIFF
--- a/lib/interactive/control-router.js
+++ b/lib/interactive/control-router.js
@@ -19,45 +19,37 @@ function controlRouter(mouseevent, mixerControls, mixerControl, gameJson, inputE
       username = participant.username;
     }
 
-    // Log input event.
-    //console.log(inputEvent);
-    //console.log(participant);
-    //console.log(control);
+    // Create request wrapper (instead of having to pass in a ton of args)
+    var processEffectsRequest = {
+        effects: effects,
+        firebot: firebot,
+        participant: participant,
+        control: control,
+        isManual: false
+    } 
 
     // Check to see if this is a mouse down or mouse up event.
     if(mouseevent == "mousedown"){
         // Mouse Down event called.
 
-        // Cooldown Button or Buttons
-        cooldown.router(mixerControls, mixerControl, firebot, control, function(response){
-                  
-          // Throw this information into the moderation panel.
-          renderWindow.webContents.send('eventlog', {username: participant.username, event: "pressed the "+controlID+" button."});
+        // Make sure cooldown is processed.
+        cooldown.router(mixerControls, mixerControl, firebot, control)
+        .then((res) => autoPlay(processEffectsRequest))
+        .then(() => {
+            // Charge sparks for the button that was pressed.
+            if (inputEvent.transactionID) {
+                // This will charge the user.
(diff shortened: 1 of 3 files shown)


lib/interactive/control-router.js
```diff
@@ -19,45 +19,37 @@ function controlRouter(mouseevent, mixerControls, mixerControl, gameJson, inputE
...
-          // Run the effects
-          effectRunner.processEffects(processEffectsRequest)
-            .then(function() {
-              // This is called after the effects are done running. 
-            });
-
-        }) // end cooldown function
+        // Make sure cooldown is processed.
+        cooldown.router(mixerControls, mixerControl, firebot, control)
+        .then((res) => autoPlay(processEffectsRequest))
```
ebiggz: @Firebottle This should work fine, but I am unclear of intended behavior. Do you want to deduct sparks after all the button effects have finished? The way you have a .then after calling the autoPlay function seems to imply that, however this setup doesn't technically wait for the effects to finish (if there is delays, anyway). 

If you don't wish to wait for effects to finish (What makes sense to me, deduction should happen immediately in my eyes), then you just could do:
```javascript
cooldown.router(mixerControls, mixerControl, firebot, control)
    .then((res) => { 
         // Start effect execution
         autoPlay(processEffectsRequest);

         // Charge sparks for the button that was pressed.
         if (inputEvent.transactionID) {
             // This will charge the user.
             mixerInteractive.sparkTransaction(inputEvent.transactionID);
         };
    })
```
That will make it more clear that you don't necessarily expect autoPlay to finish and so this removes the extra 'then'.

If you do want to wait for effects to finish, however, then you will need autoPlay to return the effect-runner promise.
```javascript
function autoPlay(processEffectsRequest){
     // Run the effects
     return effectRunner.processEffects(processEffectsRequest);
 }
```
then you can do :
```javascript
cooldown.router(mixerControls, mixerControl, firebot, control)
    .then((res) => { 
         // Start effect execution
         return autoPlay(processEffectsRequest);
    })
    .then(() => { 
         // Charge sparks for the button that was pressed.
         if (inputEvent.transactionID) {
             // This will charge the user.
             mixerInteractive.sparkTransaction(inputEvent.transactionID);
         };
    })
```
Ultimately this is just a style change, not a functional change, so at the end of the day it's personal preference. Whatever you prefer! Looks like it will work great as is.

kateract:
My only comment would be that this might be something that we want to consider implementing in the interactive-node repository, maybe add an optional parameter to setCooldown() of overwriteExisting and have the default behavior to keep the old, longer cooldown.

Firebottle:
Yep, I don't intend to wait until the effects are all done since it could take an unknown amount of time. Ideally I think that'd be good to make sure all effects triggered something (the person got what they're paying for) before charging them, but with the delay addition I don't think that's a good way to handle it.

Will make the changes you suggested for clarity.

PR #207, opened 2017-06-23
Change Group and Change Scene Effects

Firebottle:
This adds in the change group and change scene effects.


gui/app/services/effectHelperService.js
```diff
@@ -102,6 +102,79 @@
...
+            $scope.viewerGroups = [];
+            var groups = groupsService.getViewerGroups();
+            for (group of groups){
+              $scope.viewerGroups.push(group.groupName);
+            }
+
+            // This is run each time a group checkbox is clicked or unclicked.
+            // This will build an array of currently selected groups to be saved to JSON.
+            $scope.groupArray = function(group){
+              if($scope.effect.groups !== undefined){
```
ebiggz: I recently looked this up and learned that it's best practice to do `variable != null` since null and undefined are two different things. Doing `!== undefined` only checks for undefined but doing `!= null` will convert "undefined" to null so you cover both possibilities. While it's unlikely that this will be null, I don't think it hurts to be safe.

gui/app/services/effectHelperService.js
```diff
@@ -102,6 +102,79 @@
...
+
+              // Set new scope var.
+              $scope.effect.groups = groupArray;
+              console.log(groupArray)
+            }
+
+            // This uses the board service to get a list of scenes for the current board.
+            $scope.getScenesForSelectedBoard = function (){
+              var scenes = boardService.getScenesForSelectedBoard();
+              return scenes;
```
ebiggz: This is a small thing, but if you wanted to make this more concise, you could either write it as:
```javascript
$scope.getScenesForSelectedBoard = boardService.getScenesForSelectedBoard;
```
or
```javascript
$scope.getScenesForSelectedBoard = function (){
     return boardService.getScenesForSelectedBoard();
}
```

gui/app/services/effectHelperService.js
```diff
@@ -102,6 +102,79 @@
...
+            // This checks if an item is in the effect.group array and returns true.
+            // This allows us to check boxes when loading up this button effect.
+            $scope.groupCheckboxer = function (group){
+              if($scope.effect.groups.indexOf(group) != -1){
+                return true;
+              } else {
+                return false;
+              }
+            }
+
```
ebiggz: You could shorten this to
```javascript
$scope.groupCheckboxer = function (group){
     return $scope.effect.groups.indexOf(group) != -1;
}
```

PR #209, opened 2017-06-24
New Button Effects

Firebottle:
Celebration, cooldown, custom scripts, dice, html button effects added. 

**BREAKING**: Scripts folder moved inside of user settings to better fit in with the overall folder structure.

### DIFF
--- a/gui/app/services/boardService.js
+++ b/gui/app/services/boardService.js
@@ -116,6 +116,15 @@
         return scenes;
     }
     
+    service.getControlIdsForSelectedBoard = function (){
+        var board = service.getLastUsedBoard();
+        var controls = [];
+        if (board != null) {
+          controls = _.keys(board.controls);
+        }
+        return controls;
+    }
+
     /**
     *  Private helper methods
     */
--- a/gui/app/services/boardService.js
+++ b/gui/app/services/boardService.js
@@ -111,7 +111,7 @@
         var board = service.getLastUsedBoard();
         var scenes = [];
         if (board != null) {
-          scenes = _.keys(board.scenes);
+          scenes = Object.keys(board.scenes);
         }
         return scenes;
     }
@@ -120,7 +120,7 @@
         var board = service.getLastUsedBoard();
         var controls = [];
         if (board != null) {
-          controls = _.keys(board.controls);
+          controls = Object.keys(board.controls);
         }
         return controls;
     }
--- a/gui/app/services/effectHelperService.js
+++ b/gui/app/services/effectHelperService.js
@@ -166,6 +166,11 @@
             $scope.buttonCheckboxer = function (list, item){
               return service.checkSavedArray(list, item);         
             }
+
+            // Uncheck all checkboxes.
+            $scope.uncheckAll = function() {
+                $scope.effect.buttons = [];
+            }
             
           };
           break;
--- a/gui/app/services/effectHelperService.js
+++ b/gui/app/services/effectHelperService.js
@@ -198,6 +198,112 @@
 
           };
           break;
+        
+        case EffectType.GAME_CONTROL:
+          controller = ($scope) => {
+
+            $scope.validControls = [
+              "a",
+              "b",
+              "c",
+              "d",
+              "e",
+              "f",
+              "g",
+              "h",
+              "i",
+              "j",
+              "k",
+              "l",
+              "m",
+              "n",
+              "o",
+              "p",
+              "q",
+              "r",
+              "s",
+              "t",
+              "u",
+              "v",
+              "w",
+              "x",
+              "y",
+              "z",
+              "1",
+              "2",
+              "3",
+              "4",
+              "5",
+              "6",
+              "7",
+              "8",
+              "9",
+              "0",
+              "backspace",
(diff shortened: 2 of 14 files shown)


gui/app/services/boardService.js
```diff
@@ -116,6 +116,15 @@
         return scenes;
     }
     
+    service.getControlIdsForSelectedBoard = function (){
+        var board = service.getLastUsedBoard();
+        var controls = [];
+        if (board != null) {
+          controls = _.keys(board.controls);
```
ebiggz: Per Probable's recommendation, I am going to actually remove underscore.js since the newer version of Javascript supports most of the functions that underscore. So anything going forward should use native implementations.

In this case, instead of `_.keys()` you can do `Object.keys(board.controls)`

gui/app/services/effectHelperService.js
```diff
@@ -163,16 +143,62 @@
...
+            $scope.scriptArray = fs.readdirSync('./user-settings/scripts');
+
+            // Grab files in folder on refresh click.
+            $scope.getNewScripts = function (){
+              $scope.scriptArray = fs.readdirSync('./user-settings/scripts');
+            }
+
+            // Open script folder on click.
+            $scope.openScriptsFolder = function(){
+              ipcRenderer.send('openScriptsFolder');
```
ebiggz: At the moment, this is entirely sufficient but we eventually might want to move event sending to the listenerService (and rename that to the eventService) so we have once consolidated place for sending and listening to events.

gui/app/templates/interactive/effect-options/cooldown.html
```diff
@@ -0,0 +1,24 @@
...
+    <div class="cooldown-group-buttons">
+        <ul class="cooldown-group-buttons-list">
+            <li ng-repeat="button in boardButtons">
+                <input ng-click="buttonArray(effect.buttons,button)" ng-checked="buttonCheckboxer(effect.buttons,button)" type="checkbox" aria-label="..." >
+                <span>{{button}}</span>
+            </li>
+        </ul>
+    </div>
+    <div class="cooldown-group-button-reset">
+        <button class="btn btn-warning">Uncheck all</button>
```
ebiggz: This button currently wont do anything. We'll need to add logic for it.

gui/app/templates/interactive/effect-options/custom-script.html
```diff
@@ -0,0 +1,20 @@
+<div class="effect-setting-container">
+    <div class="effect-specific-title"><h4>Which script would you like to use?</h4></div>
+    <div class="effect-info alert alert-info">
+        Place scripts in the <a id="scriptFolderBtn" href="#" ng-click="openScriptsFolder()">scripts folder</a> of the root Firebot directory, then refresh the dropdown.
```
ebiggz: We will want to update the copy for this since it still refers to the "root" directory. Maybe something like:

"Place scripts in the scripts folder in Firebot's /user-settings directory, then refresh the dropdown."

For the <a> tag, we will probably want to add an underline to this link since this the blue info background makes it hard to tell its clickable: `style="text-decoration:underline"`

gui/app/templates/interactive/effect-options/custom-script.html
```diff
@@ -0,0 +1,20 @@
...
+    </div>
+    <div class="btn-group">
+        <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+            <span class="script-type">{{effect.scriptName ? effect.scriptName : 'Pick one'}}</span> <span class="caret"></span>
+        </button>
+        <a ng-click="getNewScripts()" id="refreshScriptList" href="#" style="padding-left:5px;height:100%;"><i class="fa fa-refresh" id="refreshIcon" style="margin-top:10px;" aria-hidden="true"></i></a>
+        <ul class="dropdown-menu script-dropdown">
+            <li ng-repeat="script in scriptArray" ng-click="effect.scriptName = script">
+                <a href>{{script}}</a>
+            </li>
```
ebiggz: We could add a conditional \<li> element here when there is no scripts found, just for a better user experience. 

````
<li ng-show="scriptArray.length == 0" class="muted">
      <a href>No scripts found.</a>
</li>
````
"muted" is a css class I just added (so you may need to do a pull to see it), but it just sets the element to have an opacity of 0.5, giving it a softer look.

File: gui/app/templates/interactive/effect-options/cooldown.html (hunk 3 above)
Comment (Firebottle) [inline gui/app/templates/interactive/effect-options/cooldown.html]: I took a look at this and I'm not sure what the best way to handle it is yet. All the examples I found put ng-model="all" on the uncheck button and then ng-checked="all" on the inputs. But, if I do this then I'll need to replace the function currently there which checks the correct boxes on load.

Recommendations?

Comment (ebiggz) [inline gui/app/templates/interactive/effect-options/cooldown.html]: So what we need to do is empty out the list of buttons that are checked.

```
$scope.uncheckAll = function() {
    $scope.effect.buttons = [];
}
```
Then add `ng-click="uncheckAll()"` to the button.

PR #210, opened 2017-06-25
Tab the settings page to organize it a bit.

Firebottle:
I added tabs to the settings page to try to organize it a bit. There are now tabs for "General", "Interactive", and "Overlay". This should allow us to add even more settings while making it easier for people to find what they're looking for in there.


ebiggz:
This looks great!

PR #211, opened 2017-06-26
Updates page

Firebottle:
This adds in the updates page. It works the same as the current updates page, except if there is an update available it will show an alert at the top with instructions on how to update.

We could probably also use an indicator somewhere in the app that an update is available so they don't need to click on the tab to find out.


gui/app/services/updatesService.js
```diff
@@ -0,0 +1,89 @@
...
+ angular
+  .module('firebotApp')
+  .factory('updatesService', function (settingsService) {
+    // factory/service object
+    var service = {}
+
+    // Update Checker
+    // This checks for updates.
+    service.updateCheck = function(){
+        return new Promise((resolve, reject) => {
```
ebiggz: I recently learned how to make promises work better in Angular. You need to use Angular's promise implementation, which is called the $q service (odd name). Nice thing is you can just toss an existing Promise into it and it works (and then you wont need $scope.apply in the controller.

First inject $q into the updatesService
```javascript
.factory('updatesService', function (settingsService, $q) {
```
Then wrap the promise in $q.when();
```javscript
return $q.when(
    new Promise((resolve, reject) => {
        //logic 
}));
```
And now you should be able to remove the $scope.apply in the updates controller.

gui/app/controllers/updatesController.js
```diff
@@ -4,7 +4,20 @@
...
+      updatesService.updateCheck()
+        .then(update => {
+          $scope.updates = update;
+
+          // This is required by angular to strip out bad HTML that could cause issues.
+          $scope.updates.gitNotes = $sce.trustAsHtml($scope.updates.gitNotes)
+
+          // Need this as apparently angular doesn't handle promises well and you have to
+          // manually force it to update after variables change.
+          $scope.$apply();
```
ebiggz: Take a look at the comment I leave for the updatesService to see how you can avoid using $scope.$apply.

The reason regular promises don't cause Angular to update is because Angular uses something called "dirty updating" to check if variables have changed. Look it up if you're interested, but in our case we just need to use Angular's implementation of the Promise, which properly triggers scope updates.

gui/app/services/updatesService.js
Firebottle: So, I tried this change in the service file and couldn't get it to play nice. Wrapping the entire promise in $q.when seems to break things, so I tried wrapping only the resolve and that made no difference.

So, I moved the $q.when to the controller and just wrapped the function call. That seems to have done the trick and I was able to remove scope.apply.

PR #212, opened 2017-06-26
Rewrote the push of controls to board config to lower the amount of file operations.

ThePerry-2015:


gui/app/services/boardService.js
```diff
@@ -264,7 +264,7 @@
                         }
                         // Prepare to push to db
                         var control = {
-                            controlID: controlID,
+                            controlId: controlID,
```
ebiggz: 😆

PR #221, opened 2017-06-28
Added Video overlay options

ThePerry-2015:


ebiggz:
This is awesome. Looks really good.

PR #250, opened 2017-07-03
Error fixes and board name checks

### LINKED ISSUE #191: "Failed to Press Button" with Game Control
Most of the time a Game Control button is pressed, the entire board disconnects with the message "Failed to Press Button" or similar.

Firebottle:
Helps fix #191 

This will now check the name of the board for periods. If one exists it will abort loading the board and ask the user to change the name.

Also fixed the error log code that was left in board service from the last version. It was just throwing console errors before, and now it's pulling from utility service.

### DIFF
--- a/gui/app/services/boardService.js
+++ b/gui/app/services/boardService.js
@@ -8,7 +8,7 @@
  
  angular
   .module('firebotApp')
-  .factory('boardService', function ($http, $q, settingsService, $rootScope) {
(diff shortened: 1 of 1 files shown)


ebiggz:
Looks good. I wonder why a period is causing files to split into two like that. I have a feeling it’s a bug or limitation with JsonDb.

PR #251, opened 2017-07-03
Cleaning up some overlay code. Adding notification.

Firebottle:
Cleaning up overlay code a bit, converting it to scss, and adding notification to help troubleshoot connection issues.

- Converted overlay over to scss.
- Simplified positioning styles so that images, videos, and other effects can use them without duplicating css.
- Added in notification popup that will show when trying to connect to Firebot. Should help troubleshoot connection problems and warn streamers of issues.
- Separated out effect code into different js files.

Notification just shows in the top left of the overlay until Firebot connects. The 8080 is the port it's trying to connect to so we can troubleshoot that easier.
image


ebiggz:
This looks great!

PR #252, opened 2017-07-03
Create firebottle app

RicardoArturo:
I know they are very busy perfecting the firebottle program.

But I would like to know if they could in the future create their .exe as an APP.


Firebottle:
Did you mean to open an issue?

PR #258, opened 2017-07-04
Calculate 16 by 9 for videos.

Firebottle:
This automatically calculates 16 by 9 for videos and fills out the opposite size field.

### DIFF
--- a/gui/app/services/effectHelperService.js
+++ b/gui/app/services/effectHelperService.js
@@ -92,16 +92,27 @@
               $scope.effect.videoType = type;
             }
 
+            // Force ratio toggle
+            $scope.forceRatio = true;
+            $scope.forceRatioToggle = function(){
+              if($scope.forceRatio === true){
+                $scope.forceRatio = false;
+              } else {
+                $scope.forceRatio = true;
+              }
+            }
+
             // Calculate 16:9
             // This checks to see which field the user is filling out, and then adjust the other field so it's always 16:9.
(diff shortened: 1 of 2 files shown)


gui/app/services/effectHelperService.js
```diff
@@ -86,6 +86,26 @@
...
+            }
+
+            // Calculate 16:9
+            // This checks to see which field the user is filling out, and then adjust the other field so it's always 16:9.
+            $scope.calculateSize = function(widthOrHeight, size) {
+              if(size !== ""){
+                if(widthOrHeight == "Width"){
+                  $scope.effect.height = Math.round((size/9)*16);
+                } else {
+                  $scope.effect.width = Math.round((size/16)*9);
```
ebiggz: These equations are slightly off, here's what they should be:
```js
if(widthOrHeight == "Width"){
    $scope.effect.height = Math.round((size/16)*9);
 } else {
    $scope.effect.width = Math.round((size*16)/9);
 }
```

Firebottle:
Made the changes you suggested. Keeping the styles inline until the theme switcher thing is moved over otherwise we'll get merge conflicts. 

Do we want to save the checkbox setting to the effect json? Right now you can toggle it off but next time you visit that effect it'll be on again.

ebiggz:
We could save the state of the checkbox per saved video effect, but I dont think its a huge deal. Either way, its super easy to implement if we want to.

PR #260, opened 2017-07-04
Theme Switcher

Firebottle:
This adds theme switching and sets up the sass for easily adding more themes in the future.

- Added ability to switch between different themes.
- Cleaned up and removed some unneeded css and css files from previous versions.
- Built variable template for light and dark theme that can be copy/pasted to create more themes if needed.

I had to change a ton of stuff for this one. So there might be some styles that need to be tweaked a bit.


ebiggz:
I've pushed some style changes to this branch to fix some issues and make sure the style matches better to how it was before. Only real difference is that modals are still a tad darker than they used to be. If you guys like this, then we are good to go.

PR #287, opened 2017-07-23
Added version string + package script

ThePerry-2015:
Added a version string to control the name of the app shown on the desktop shortcut to the pack:win64 line plus copying necessary files over into the dist folder after building the Electron Package.
Added a script line for auto build electron package as well as installer package.


ProbablePrime:
You should move those switch statements to a config file. 

I see a grunt file so: https://github.com/electron-userland/electron-packager/blob/master/docs/api.md should be all you need

PR #368, opened 2017-09-24
Perry backup function

ThePerry-2015:
Adding backup routine for manually taking backups plus options to make automatic backups on application exit and/or on updates as well as an option to keep all backups or overwriting the backup file each time.


ebiggz:
This is awesome.

PR #406, opened 2017-10-17
Voice channel (discord)

RicardoArturo:
Create a (Play) button to place in Mixer and listen (X) Discord channel.

When you broadcast your screen and you are playing with other friends that occupy your Mixer channel, they are heard twice.

I can divide the output of adudio and my colleagues who play in my channel are heard in my headphones while I only transmit what goes on the screen.

But the problem with that is that viewers can not listen to those who are playing with me through Mixer.

The button would be so that the person that transmits only transmits the audio of the game and the spectators use the (Play) to be able to listen to the other people who are playing with me by Discord.

I hope you understand and see it useful and above all can be done :)

### DIFF
--- a/api/v1/controllers/effectsApiController.js
+++ b/api/v1/controllers/effectsApiController.js
@@ -81,13 +81,13 @@ exports.runEffects = function(req, res) {
     var boardJson = dbControls.getData('/');  
     
     var processEffectsRequest = {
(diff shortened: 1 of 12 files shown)


Firebottle:
Use the issues tab instead of the pull request tab for feature requests.

RicardoArturo:
ok

PR #496, opened 2018-01-30
Added password to script parameter options

kateract:
Added "password" option to custom script parameter fields.


ebiggz:
Ooo never even thought of this, awesome!

PR #578, opened 2018-03-13
Translations Phase 1

### LINKED ISSUE #545: Translations
Implement a system for translations of all the text in the application. 

You don't need to personally go through the entire app yourself, just implement an easy to use system and we'll catch everything as we rewrite. Reason being, a lot of the app will be rewritten before v5 is released and we'd just be overwriting your work.

kateract:
Added angular-translate
Moved some strings from sidebar and buttons tab to /gui/app/lang/locale-en.json
Other languages can be sourced via files or via web service using angular-translate-loader-url

Also removed confusing connection tooltip code in sidebar that wasn't used to generate tooltips.
Resolves #545

### DIFF
--- a/gui/app/controllers/mainController.js
+++ b/gui/app/controllers/mainController.js
@@ -7,7 +7,7 @@
     let app = angular
         .module('firebotApp',
             ['ngAnimate', 'ngRoute', 'ui.bootstrap', 'rzModule', 'ui.select', 'ngSanitize', 'ui.select', 'ui.sortable',
-                'luegg.directives', 'summernote']);
+                'luegg.directives', 'summernote', 'pascalprecht.translate']);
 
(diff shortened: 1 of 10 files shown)


ebiggz:
@kateract Everything looks good other than removing the tooltip code on line 79 of sidebar.js. It's definitely being used. It's used when the sidebar is contracted (aka when it slims down after you click the arrow in the top corner), as the user has no way to see all connection details when the toolbar is in that state. The tooltip provides that info to the user without them having to expand the sidebar again.

image

If you feel there's an issue with its implementation in code, I'm happy to take suggestions to make it better and/or less confusing, but I feel the end result (providing user with connection details when the sidebar is contracted) is valid and shouldn't be completely removed.

PR #611, opened 2018-05-04
Update htmlHandler.js

dbkirk4211:
If no CSS class name is provided or if the class isn't found, the element that is added gets removed. Also a default time of 10 seconds is set for the show time in case the field is blank.


dbkirk4211:
No problem!

PR #646, opened 2019-03-02
Add pack:linux to package.json

Teteros:
Script entry for linux package, similar to mac/win64
Executable via `npm run pack:linux`

Note: You might want also want to bump electron-packager so it works with node 10+
https://github.com/electron-userland/electron-packager/issues/831#issuecomment-450899748


Firebottle:
I'm going to go ahead and close this one. @Teteros so sorry for not getting back to you earlier. This one sort of slipped past us as we've been focusing on v5 for the most part, which should be out fairly soon.

Were you able to test this at all? Does the firebot build in linux work as expected?

PR #663, opened 2020-02-01
Allow reading and deleting specific lines in files.

LastMageTV:
Changes:

- new variable to count the number of lines in a file.
- update write file effect to allow deleting specific lines OR everything in the file.
- update read file variable to allow for reading a specific line from the file.

Notes:
- All of these changes filter our blank lines in a file. So, if you have 10 lines, but 5 of them are just empty. Then the line counter will report that there are 5 lines in the file.

### DIFF
--- a/backend/effects/builtin/fileWriter.js
+++ b/backend/effects/builtin/fileWriter.js
@@ -18,7 +18,7 @@ const fileWriter = {
     definition: {
         id: "firebot:filewriter",
         name: "Write To File",
-        description: "Write some text to a file.",
+        description: "Write or delete some text in a file.",
         tags: ["Built in"],
         dependencies: [],
         triggers: effectModels.buildEffectTriggersObject(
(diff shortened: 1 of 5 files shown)


backend/effects/builtin/fileWriter.js
```diff
@@ -40,22 +37,30 @@ const fileWriter = {
...
-            </label>
-            <label class="control-fb control--radio">Append
-                <input type="radio" ng-model="effect.writeMode" value="append"/>
-                <div class="control__indicator"></div>
-            </label>
-        </div>
-    </eos-container>
+        <eos-container header="Text or Line Number(s)" pad-top="true" ng-if="effect.writeMode === 'replace' || effect.writeMode === 'append' || effect.writeMode === 'delete'">
+            <input ng-model="effect.text" type="text" class="form-control" id="chat-text-setting" placeholder="Enter text" replace-variables>
+        </eos-container>
```
ebiggz: Hopefully its not too nit-picky, but I'd like to try to avoid having one text field be used for two completely different purposes. It makes things a little ambiguous for the user but very ambiguous in the code. Having separate text fields would also let us only show variables that only output numbers in the Line Number(s) field.

Proposed change:
```html
<eos-container header="Text" pad-top="true" ng-if="effect.writeMode === 'replace' || effect.writeMode === 'append'">
    <input ng-model="effect.text" type="text" class="form-control" id="chat-text-setting" placeholder="Enter text" replace-variables>
</eos-container>

<eos-container header="Line Number(s)" pad-top="true" ng-if="effect.writeMode === 'delete'">
    <p class="muted">Enter a line number or list of line numbers (separated by commas) to delete.</p>
    <input ng-model="effect.lineNumbers" type="text" class="form-control" id="chat-line-numbers-setting" placeholder="Enter line number(s)" replace-variables="number">
</eos-container>
```
Note that I did` replace-variables="number"`. This will make the variables menu only show variables that have an output type of "number". 

Also note that you will have to update `fileWriterProcessor.js` to account for the now separated `lineNumbers` property.

backend/variables/builtin/fileLineCount.js
```diff
@@ -0,0 +1,38 @@
...
+
+        try {
+            let contents = fs.readFileSync(filePath, "utf8");
+            let lines = contents.split('\n');
+            let filteredLines = lines.filter(function (line) {
+                if (line != null && line.trim() !== "") {
+                    return line;
+                }
+            });
+            return filteredLines.length;
```
ebiggz: You could shorten this to:
```js
let lines = contents
    .split('\n')
    .filter(l => l != null && l.trim() !== "");

return lines.length;
```

backend/variables/builtin/readFile.js
```diff
@@ -12,19 +12,36 @@ const model = {
...
+
+            if (randomLine === "true" || !isNaN(randomLine)) {
                 let lines = contents.replace(/\r\n/g, "\n").split("\n");
-                let randIndex = util.getRandomInt(0, lines.length - 1);
-                return lines[randIndex];
+                filteredLines = lines.filter(function (line) {
+                    if (line != null && line.trim() !== "") {
+                        return line;
+                    }
+                });
```
ebiggz: You could shorten this to:
```js
filteredLines = lines.filter(l => l != null && l.trim() !== "");
```

backend/variables/builtin/readFile.js
```diff
@@ -12,19 +12,36 @@ const model = {
...
     },
     evaluator: (_, filePath, randomLine) => {
 
         if (filePath === null || !filePath.endsWith(".txt")) return "[File Path Error]";
         try {
             let contents = fs.readFileSync(filePath, "utf8");
-            if (randomLine) {
+            let filteredLines = [];
+
+            if (randomLine === "true" || !isNaN(randomLine)) {
```
ebiggz: If the user just inputs `true` for the second variable, it will show up as a boolean, not a string, because the variable system is suppah smart (thanks Reject!), so we need to account for that.

```js
let shouldReadRandomLine = randomLine === true || randomLine === "true";
if (shouldReadRandomLine || !isNaN(randomLine)) {
```

backend/variables/builtin/readFile.js
```diff
@@ -12,19 +12,36 @@ const model = {
...
-                return lines[randIndex];
+                filteredLines = lines.filter(function (line) {
+                    if (line != null && line.trim() !== "") {
+                        return line;
+                    }
+                });
+            }
+
+            // Get line at random.
+            if (randomLine === "true") {
```
ebiggz: Use new `shouldReadRandomLine` variable suggested above:
```suggestion
            if (shouldReadRandomLine ) {
```

LastMageTV:
@ebiggz Davemane just came up with a cool thing! This would allow people to run giveaways now without a full featured giveaway system.

Command 1 (!startgiveaway)
- Clears out the file entirely.
- Spits out chat message that says giveaway is started.

Command 2 (!joingiveaway)
- Cooldown this button per user longer that join giveaway period.
- Write username to text file.
- Whisper them that they joined.

Command 3 (!endgiveaway)
- Get random line from the file.
- Spit out the username and winner text in chat.

PR #669, opened 2020-02-10
Active Chatter and Currency Adjustments

LastMageTV:
Changes:
- Set up active chatter as a "Firebot Role".
- Add Firebot Roles to the bonus currency area on the create/edit currency modal.
- Add Firebot Roles to the currency effect "role" option.
- Show Firebot Roles on the Viewer Roles screen.
- Add "Set" option to currency effect that will just set the currency to a specific value.
- Create isActiveChatterUsername function to see if a specific person is active.
- Cleaned up areas to use the isActiveChatterUsername function.

Notes:
- We already have "Active Viewers" as a filter. So, I left Firebot Roles out of the "permissions" filter. If we add more Firebot Roles later on that we want to show in the permissions area, we might need to figure out that problem. Probably "Active Viewers" would just map to that Firebot Role or something.

### DIFF
--- a/backend/database/currencyDatabase.js
+++ b/backend/database/currencyDatabase.js
@@ -7,6 +7,7 @@ const { settings } = require("../common/settings-access.js");
 const channelAccess = require("../common/channel-access");
 const customRolesManager = require("../roles/custom-roles-manager");
 const mixerRolesManager = require("../../shared/mixer-roles");
+const firebotRolesManager = require("../roles/firebot-roles-manager");
 
 let currencyCache = {};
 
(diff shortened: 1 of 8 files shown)


backend/database/currencyDatabase.js
```diff
@@ -34,26 +35,39 @@ function getCurrencies() {
...
-        }
+        adjustType = adjustType.toLowerCase();
+        let newUserValue = value;
 
-        // Okay, move on and finish everything.
-        logger.debug(
-            "Currency: Adjusting " + value + " currency to " + user.username + ". " + currencyId
-        );
+        switch (adjustType) {
+        case "set":
```
LastMageTV: adjustType is new and allows us to pass in things like "set", "addPercentage", etc.. so we can calculate the new value differently.

PR #670, opened 2020-02-14
Add ability to remove users from activeViewerList.

LastMageTV:
This puts a toggle into the viewer modal that will allow people to exempt certain users from the "Active Viewers" list. This will allow people to exclude their bots from things like the $randomActiveViewer variables.

If we decide to put in a separate $randomActiveMixplay variable for mixplay connections, this same toggle could exempt from that as well.

### DIFF
--- a/backend/chat/active-chatters.js
+++ b/backend/chat/active-chatters.js
@@ -25,18 +25,17 @@ function isUsernameActiveChatter(username) {
     return false;
 }
 
-function addOrUpdateActiveChatter(user) {
+async function addOrUpdateActiveChatter(user) {
     if (!activeUserListStatus) {
         return;
     }
 
     // Stop early if user shouldn't be in active chatter list.
-    userDatabase.getUserByUsername(user.user_name).then(userDB => {
-        if (userDB.disableActiveUserList) {
-            logger.debug(userDB.username + " is set to not join the active viewer list.");
-            return;
-        }
-    });
+    let userDB = await userDatabase.getUserByUsername(user.user_name);
+    if (userDB.disableActiveUserList) {
+        logger.debug(userDB.username + " is set to not join the active viewer list.");
+        return;
+    }
 
     // User should be okay, add them!
     let date = new Date;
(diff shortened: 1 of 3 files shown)


backend/chat/active-chatters.js
```diff
@@ -29,6 +30,15 @@ function addOrUpdateActiveChatter(user) {
         return;
     }
 
+    // Stop early if user shouldn't be in active chatter list.
+    userDatabase.getUserByUsername(user.user_name).then(userDB => {
+        if (userDB.disableActiveUserList) {
+            logger.debug(userDB.username + " is set to not join the active viewer list.");
+            return;
```
ebiggz: I think there is a bug here. The user will still get added to the active list because when you `return` here, you are only returning in the `then()` function, not the outer `addOrUpdateActiveChatter` function. So execution of `addOrUpdateActiveChatter` isn't stopped. 

I would suggest making `addOrUpdateActiveChatter` async:
```js
asyn function addOrUpdateActiveChatter(user) {
```

and then await the `getUser` call:
```js
let firebotUser = await userDatabase.getUserByUsername(user.user_name);
if (firebotUser != null && firebotUser.disableActiveUserList) {
            logger.debug(firebotUser.username + " is set to not join the active viewer list.");
            return;
}
```

gui/app/directives/modals/viewers/viewerDetailsModal.js
```diff
@@ -376,6 +383,14 @@
                     });
                 };
 
+                $ctrl.disableActiveUserList = () => {
```
ebiggz: Update this to `disableActiveUserListChange` to keep in line with the naming convention of the other similar function `disableAutoStatAccuralChange `

backend/chat/active-chatters.js
ebiggz: Going forward I think we should default to using `async/await` on promises vs doing `.then()` chaining. I think async/await is easier to read (especially compared to being chained like three .then()s deep) and it inherently helps avoid asynchronous bugs like this. 

I know `async/await` stuff is newer and a little weird at first but perhaps after a few repetitions, it'll come more naturally.

PR #673, opened 2020-02-17
V5 mixplay active users

LastMageTV:
Changes:
- Keep list of active mixplay users on backend.
- Add randomActiveMixplayUser variable
- Add randomMixplayUser variable
- Add activeMixplayUserCount variable
- Add settings to turn active mixplay list on or off and adjust inactive timeout.
- Move active-chatters file and active-mixplay-users files to the roles folder to keep role specific logic together.

Tests:
- Tested to make sure tagging a person in the viewer db to keep them off active lists works on this as well.
- Tested all three new variables.
- Tested to make sure anonymous users are not placed on the active user list. (Would be pointless for the vast majority of uses).

Notes:
- This moves the active-chatters file, so it could have impact on old active chatters variable. But, I believe I got everything pointing to the correct new location.

### DIFF
--- a/backend/common/settings-access.js
+++ b/backend/common/settings-access.js
@@ -184,6 +184,16 @@ settings.getActiveChatUserListTimeout = function() {
     return inactiveTimer != null ? parseInt(inactiveTimer) : 10;
 };
 
+settings.getActiveMixplayUserList = function() {
+    let status = getDataFromFile("/settings/activeMixplayUsers/status");
+    return status != null ? status : "On";
+};
+
+settings.getActiveMixplayUserListTimeout = function() {
+    let inactiveTimer = getDataFromFile("/settings/activeMixplayUsers/inactiveTimer");
+    return inactiveTimer != null ? parseInt(inactiveTimer) : 10;
+};
+
 settings.sparkExemptionEnabled = function() {
     let enabled = getDataFromFile('/settings/sparkExemptionEnabled');
     return enabled != null ? enabled : false;
--- a/backend/common/settings-access.js
+++ b/backend/common/settings-access.js
@@ -174,9 +174,9 @@ settings.setMaxBackupCount = function(maxBackupCount) {
     pushDataToFile("/settings/maxBackupCount", maxBackupCount);
 };
 
-settings.getActiveChatUserList = function() {
-    let status = getDataFromFile("/settings/activeChatUsers/status");
-    return status != null ? status : "On";
+settings.getActiveChatUserListEnabled = function() {
+    let enabled = getDataFromFile("/settings/activeChatUsers/status");
+    return enabled != null ? enabled : true;
 };
 
 settings.getActiveChatUserListTimeout = function() {
@@ -184,9 +184,9 @@ settings.getActiveChatUserListTimeout = function() {
     return inactiveTimer != null ? parseInt(inactiveTimer) : 10;
 };
 
-settings.getActiveMixplayUserList = function() {
-    let status = getDataFromFile("/settings/activeMixplayUsers/status");
-    return status != null ? status : "On";
+settings.getActiveMixplayUserListEnabled = function() {
+    let enabled = getDataFromFile("/settings/activeMixplayUsers/status");
+    return enabled !== undefined ? enabled : true;
 };
 
 settings.getActiveMixplayUserListTimeout = function() {
--- a/backend/interactive/mixplay.js
+++ b/backend/interactive/mixplay.js
@@ -119,7 +119,9 @@ function addControlHandlers(controls) {
 
             controlManager.handleInput(event, sceneId, inputEvent, participant);
 
-            activeMixplayUsers.addOrUpdateActiveUser(participant);
+            if (participant.anonymous !== true) {
+                activeMixplayUsers.addOrUpdateActiveUser(participant);
+            }
         };
 
         //remove previous listener just in case one exists
--- a/backend/interactive/mixplay.js
+++ b/backend/interactive/mixplay.js
@@ -388,18 +388,9 @@ mixplayClient.state.on('participantJoin', async participant => {
             await updateParticipantWithUserData(firebotUser, participant);
         }
 
-        activeMixplayUsers.addOrUpdateActiveUser(participant.username);
-
         eventManager.triggerEvent("mixer", "user-joined-mixplay", {
             username: participant.username
         });
-    }
-});
-
-mixplayClient.state.on('participantLeave', async participant => {
-    logger.debug(`${participant.username} (${participant.sessionID}) Left`);
-    if (!participant.anonymous) {
-        activeMixplayUsers.removeLeavingUser(participant.username);
     }
 });
 
--- a/backend/interactive/mixplay.js
+++ b/backend/interactive/mixplay.js
@@ -118,10 +118,6 @@ function addControlHandlers(controls) {
             logger.debug(`Control event "${event}" for control "${inputData.controlID}" in scene "${sceneId}"`);
 
             controlManager.handleInput(event, sceneId, inputEvent, participant);
-
-            if (participant.anonymous !== true) {
-                activeMixplayUsers.addOrUpdateActiveUser(participant);
-            }
         };
 
         //remove previous listener just in case one exists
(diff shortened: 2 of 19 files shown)


backend/interactive/mixplay.js
```diff
@@ -117,6 +118,10 @@ function addControlHandlers(controls) {
             logger.debug(`Control event "${event}" for control "${inputData.controlID}" in scene "${sceneId}"`);
 
             controlManager.handleInput(event, sceneId, inputEvent, participant);
+
+            if (participant.anonymous !== true) {
```
ebiggz: these sorts of conditions can be shortened to `if (!participant.anonymous)` since we can trust that property will always be a Boolean.

backend/interactive/mixplay.js
```diff
@@ -391,6 +396,23 @@ mixplayClient.state.on('participantJoin', async participant => {
...
+        if (user != null) {
+            user = participants[user][1];
+            if (user.anonymous !== true) {
+                participantNames.push(user);
+            }
+        }
+    }
+
+    return participantNames;
+}
```
ebiggz: You could shorten this method to:
```js
function getConnectedUsernames() {
    let participants = [...mixplayClient.state.getParticipants().values()];
    return participants
        .filter(p => p != null && !p.anonymous)
        .map(p => p.username);
}
```
Note I also renamed the method to make it more clear that this only returns usernames and not entire user/participant objects when calling it elsewhere in the software.
Lastly, it does not need to be `async` as nothing in this method needs to `await` on a Promise.

backend/roles/role-managers/active-mixplay-users.js
```diff
@@ -0,0 +1,142 @@
...
+const settings = require("../../common/settings-access").settings;
+const logger = require("../../logwrapper");
+const userDatabase = require("../../database/userDatabase");
+const Mixplay = require("../../interactive/mixplay");
+
+// Active user toggle
+let activeUserListStatus = settings.getActiveMixplayUserList() === false ? false : true;
+
+// User Timeout Settings
+let userInactiveTimeSetting = settings.getActiveMixplayUserListTimeout() != null ? settings.getActiveMixplayUserListTimeout() : 10;
```
ebiggz: Looks like `settings.getActiveMixplayUserListTimeout()` already does a null check so you don't need to do it here too.

backend/common/settings-access.js
```diff
@@ -184,6 +184,16 @@ settings.getActiveChatUserListTimeout = function() {
     return inactiveTimer != null ? parseInt(inactiveTimer) : 10;
 };
 
+settings.getActiveMixplayUserList = function() {
+    let status = getDataFromFile("/settings/activeMixplayUsers/status");
+    return status != null ? status : "On";
+};
```
ebiggz: I would change this to store a boolean instead of storing a string that serves the same purpose as a boolean (on/off is the same as true/false).
 I would also update the name to `settings.getActiveMixplayUserListEnabled` so its more clear what this means elsewhere (and is more in line with other newer settings that return booleans).

```js
settings.getActiveMixplayUserListEnabled = function() {
    let enabled = getDataFromFile("/settings/activeMixplayUsers/status");
    return status !== undefined ? enabled : true;
};
```

backend/roles/role-managers/active-mixplay-users.js
```diff
@@ -0,0 +1,142 @@
+"use strict";
+
+const { ipcMain } = require("electron");
+const settings = require("../../common/settings-access").settings;
+const logger = require("../../logwrapper");
+const userDatabase = require("../../database/userDatabase");
+const Mixplay = require("../../interactive/mixplay");
+
+// Active user toggle
+let activeUserListStatus = settings.getActiveMixplayUserList() === false ? false : true;
```
ebiggz: This settings method already does a null/default value check so you dont' need to do it here. Also as it stands currently, `getActiveMixplayUserList()` actually returns a string ("On" or "Off") and not a boolean as you are checking here. See my comment where that method is defined, I suggest changing it to return a boolean to avoid confusion as has happened here.

backend/roles/role-managers/active-mixplay-users.js
```diff
@@ -0,0 +1,142 @@
...
+let cycleActiveTimer = [];
+let activeMixplayUsers = [];
+
+function isUsernameActiveUser(username) {
+    let existingUserIndex = activeMixplayUsers.findIndex((obj => obj.username === username));
+    if (existingUserIndex !== -1) {
+        return true;
+    }
+    return false;
+}
```
ebiggz: You could shorten this to: 
```js
function isUsernameActiveUser(username) {
    return activeMixplayUsers.some(u => u.username === username);
}
```

backend/roles/role-managers/active-mixplay-users.js
```diff
@@ -0,0 +1,142 @@
...
+    for (let userIndex in activeMixplayUsers) {
+        if (activeMixplayUsers[userIndex] != null) {
+            let user = activeMixplayUsers[userIndex];
+            if (user.time <= expiredTime) {
+                logger.debug(user.username + " has gone inactive on mixplay. Removing them from active mixplay list.");
+                activeMixplayUsers.splice(userIndex, 1);
+            }
+        }
+    }
+}
```
ebiggz: This can be shortened to:
```js
function clearInactiveUsers() {
    logger.debug("Clearing inactive people from active mixplay users list.");
    let expiredTime = new Date().getTime() - inactiveTimer;
    activeMixplayUsers = activeMixplayUsers.filter(u => u != null && u.time <= expiredTime);
}
```

backend/roles/role-managers/active-mixplay-users.js
```diff
@@ -0,0 +1,142 @@
...
+    }
+    activeUserListStatus = settings.getActiveMixplayUserList() === false ? false : true;
+});
+
+ipcMain.on("setActiveMixplayUserTimeout", function(event, value) {
+    logger.debug('Changing active mixplay user timeout to: ' + value);
+
+    // Make sure we have a valid value, then set it.
+    value = parseInt(inactiveTimer);
+    if (isNaN(value)) {
```
ebiggz: I would just change this to `value === NaN` since you're doing `parseInt` before the `isNaN` call anyway and `parseInt` returns the value of `NaN` if value isnt a num. `isNaN` is just a convenience method that will automatically parse a string into a number and check if its valid, you dont have to parseInt first. Since you are manually parsing first tho, its unnecessary.

gui/app/services/settingsService.js
```diff
@@ -320,6 +320,24 @@
                 pushDataToFile("/settings/activeChatUsers/inactiveTimer", inactiveTimer);
             };
 
+            service.getActiveMixplayUserList = function() {
+                let status = getDataFromFile("/settings/activeMixplayUsers/status");
+                return status != null ? status : "On";
+            };
```
ebiggz: See the comment I left on the backend version of this method (settings-access.js > getActiveMixplayUserList). I suggest making this save and return a boolean instead of a string that acts like a boolean. Its less confusing and it looks like other parts of your changes are assuming this returns a boolean anyway. 

Also rename to `getActiveMixplayUserListEnabled` for clarity.

backend/interactive/mixplay.js
LastMageTV: Oh wow, so that participants variable turns the Map into an array? I'll have to research the [...] thing.

backend/roles/role-managers/active-mixplay-users.js
LastMageTV: It seems like our current linter rules dont like value === NaN for whatever reason.

backend/interactive/mixplay.js
ebiggz: It’s called the Spread operator. You can read more about it here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax 
It’s super useful. 

And to be clear, it’s spreading `getParticipants().values()`. So it’s not spreading the Map but what `.values()` returns (an Iterable object containing the values of the map). 
Ultimately `let valuesArray = [...map.values()];` is the same as `let valuesArray = Array.from(map.values())`, it’s just the newer (and nicer imo) way.

backend/roles/role-managers/active-mixplay-users.js
ebiggz: In that case I would call isNaN like before but just move the parseInt call after it, instead of having it before.

gui/app/services/settingsService.js
ebiggz: This particular bit isn’t that important. I mainly just wanted to point out that we don’t need to call parseInt before calling isNaN. Doesn’t change much in this particular example but it can simply code a bit in other situations.

PR #719, opened 2020-02-29
Fix math cooldowns

### LINKED ISSUE #701: [Bug] Add and subtract on cooldown groups functions incorrectly
@TheLastMage via Discord:

If you use the add or subtract option in the cooldown effect on a cooldown group that is not current on it's cooldown timer, the effect will start the timer THEN add the additional amount. So, if the cooldown group is set for 20 minutes and your effect is "add 10", then the group cools down for 30 minutes. I would expect it to add 10 minutes to the current cooldown of 0.

LastMageTV:
Should fix #701 

Changes:
- Remove an extra loop.
- Return ttl immediately if it's already 0, instead of comparing with moment.

For whatever reason I had decided in my old code to do an additional loop that wasn't needed. So, for each control ID I was looping through the control IDs again. I believe this was causing cooldowns to be applied twice and cause issues. This would have only affected the "add" and "subtract" cooldowns if updating a cooldown group with multiple buttons.

### DIFF
--- a/backend/interactive/cooldown-manager.js
+++ b/backend/interactive/cooldown-manager.js
@@ -223,7 +223,7 @@ async function mathUpdate(updateData) {
             newCooldown = remainingCooldown - duration;
         }
 
-        await mixplay.updateCooldownForControls(controlIds, newCooldown);
+        await mixplay.updateCooldownForControls([controlId], newCooldown);
         updateCooldownForControlId(controlId, newCooldown / 1000);
     }
 }
--- a/backend/interactive/cooldown-manager.js
+++ b/backend/interactive/cooldown-manager.js
@@ -23,6 +23,11 @@ function cooldownIsExpired(controlId) {
 // Returns the amount of time remaining for a control id.
 function getRemainingCooldownTime(controlId) {
     let currentTtl = cooldownCache.getTtl(controlId) || 0;
+
+    if (currentTtl === 0) {
+        return 0;
+    }
+
     let diff = moment(currentTtl).diff(moment());
     if (diff < 0) {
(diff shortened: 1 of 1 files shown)


backend/interactive/cooldown-manager.js
```diff
@@ -219,9 +224,7 @@ async function mathUpdate(updateData) {
         }
 
         await mixplay.updateCooldownForControls(controlIds, newCooldown);
-        for (let controlId of controlIds) {
-            updateCooldownForControlId(controlId, newCooldown / 1000);
-        }
+        updateCooldownForControlId(controlId, newCooldown / 1000);
```
ebiggz: I am wondering if there is still a bug here...
Currently this block looks like:
```js
for (let controlId of controlIds) {
        let remainingCooldown = getRemainingCooldownTime(controlId);
        let newCooldown;

        if (updateData.type === "add") {
            newCooldown = remainingCooldown + duration;
        } else {
            newCooldown = remainingCooldown - duration;
        }

        await mixplay.updateCooldownForControls(controlIds, newCooldown);
        updateCooldownForControlId(controlId, newCooldown / 1000);
}
```
It looks like we are looping through `controlIds` and updating ALL controls each time with `await mixplay.updateCooldownForControls(controlIds, newCooldown);`... is that correct? Seems we should only need to call `updateCooldownForControls` once or even not at all since we are updating the cooldown per control with `updateCooldownForControlId(controlId, newCooldown / 1000);` afterwards.

LastMageTV: Nope, definitely unintended. Will fix it!

Issue #724, opened 2020-03-01
[Feature Request] Text value left side in roles conditional


CrowbarToolsRobot:
@TheLastMage via Discord:

We need the left side of the roles conditional to be a text field. That way you can check the roles of people other than the person interacting with the button.

PR #725, opened 2020-03-01
Implement firebot roles and allow text on left side of conditional.

LastMageTV:
Implements #724 

This updates the "viewer roles" conditional to allow for a left side value. This would allow people to do things like, "If randomTargetUser is in ROLE, then do conditional". Before the target was always the user pressing the button.

Changes:
- Add firebot roles into the conditional roles.
- Make left side a text field.
- Put fallback in place where if left side is empty, we just use the username of the person interacting with the button. This should cover people using the old version of the conditional.

### DIFF
--- a/backend/effects/builtin/conditional-effects/conditions/builtin/viewer-roles.js
+++ b/backend/effects/builtin/conditional-effects/conditions/builtin/viewer-roles.js
@@ -2,6 +2,7 @@
 
 const channelAccess = require("../../../../../common/channel-access");
 
+const firebotRolesManager = require("../../../../../roles/firebot-roles-manager");
 const customRolesManager = require("../../../../../roles/custom-roles-manager");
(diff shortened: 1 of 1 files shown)


backend/effects/builtin/conditional-effects/conditions/builtin/viewer-roles.js
```diff
@@ -44,21 +48,21 @@ module.exports = {
...
 
-        let { comparisonType, rightSideValue } = conditionSettings;
+        let { comparisonType, leftSideValue, rightSideValue } = conditionSettings;
+        let username = leftSideValue;
 
-        let username = trigger.metadata.username;
         if (username == null || username === "") {
-            return false;
+            username = trigger.metadata.username;
         }
```
LastMageTV: I believe this will cover people using the old version of the conditional. If leftSideValue is blank, then use the username of the person triggering the conditional.

Issue #744, opened 2020-03-07
[Feature Request] Follower only restriction for interactive


CrowbarToolsRobot:
@TheLastMage via Discord:

Unpack the follower only script and implement it into the bot directly as a restriction or a conditional.

PR #745, opened 2020-03-08
Follower restrictions

LastMageTV:
Implements #744 

Changes:
- New function in user-access that allows us to determine if a user is following certain streamers.
- Add follow check restriction.
- Add follow check conditional.

This will allow people to put buttons behind a list of required follows. This will be useful for people who want to cross promote other streamers by unlocking buttons and effects. It will also be useful in limiting your buttons to being follower only.

Downside:
This makes API calls. So, if a conditional or restriction is put onto something that is spammed a ton (like a chat message event) they could maybe hit the API call limit. This should cause firebot to return false on any other follow checks, but nothing should break.

### DIFF
--- a/backend/common/user-access.js
+++ b/backend/common/user-access.js
@@ -4,45 +4,57 @@ const userDb = require("../database/userDatabase");
 const accountAccess = require("../common/account-access");
 const channelAccess = require("./channel-access");
 const mixerApi = require("../api-access");
+const NodeCache = require('node-cache');
 const logger = require('../logwrapper');
 
-async function userFollowsUsers(userId, followCheckList) {
-    let userFollowsUsers = false;
+const followCache = new NodeCache({ stdTTL: 10, checkperiod: 10 });
 
-    for (let streamer of followCheckList) {
-        let streamerData = null;
+async function userFollowsChannels(userId, channelNamesOrIds) {
+    let userfollowsAllChannels = true;
 
-        if (isNaN(streamer)) {
-            try {
-                streamerData = await mixerApi.get(`channels/${streamer}?fields=id`, "v1", false, false);
-                streamer = streamerData.id;
-            } catch (err) {
-                logger.debug("Unable to get streamer data for follow check.", err);
-                userFollowsUsers = false;
-                break;
+    for (let channelNameOrId of channelNamesOrIds) {
+        let userFollowsChannel = false;
+
+        // check cache first
+        let cachedFollow = followCache.get(`${userId}:${channelNameOrId}`);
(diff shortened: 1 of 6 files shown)


backend/common/user-access.js
```diff
@@ -4,24 +4,62 @@ const userDb = require("../database/userDatabase");
...
+            userFollowsUsers = true;
+            continue;
+        }
+
+        userFollowsUsers = false;
+        break;
+    }
+
+    return userFollowsUsers;
+}
```
ebiggz: As you mentioned in the description, this could be spammed which has implications for the api call limit but also for performance (especially since **two** api calls happen in the most common flow: one to get channel id, one to get relationship data). I think we should use Node Cache with a 10 to 15 sec timeout to help alleviate this. We can also simplify the above function to make it a little more readable. 
Here is an updated example you should be able to drop right in (note that I updated the function name so you'll need to update instances that call it):
```js
const followCache = new NodeCache({ stdTTL: 10, checkperiod: 10 });

async function userFollowsChannels(userId, channelNamesOrIds) {
    let userfollowsAllChannels = true;

    for (let channelNameOrId of channelNamesOrIds) {

        let userFollowsChannel = false;

        // check cache first
        let cachedFollow = followCache.get(`${userId}:${channelNameOrId}`);
        if (cachedFollow !== undefined) {
            userFollowsChannel = cachedFollow;
        } else {
            // fetch relationship data

            let channelId;
            // fetch channel id if needed
            if (isNaN(channelNameOrId)) {
                const ids = await channelAccess.getIdsFromUsername(channelNameOrId);
                if (ids == null) {
                    // channel doesnt exist
                    continue;
                }
                channelId = ids.channelId;
            } else {
                channelId = channelNameOrId;
            }

            let userRelationshipData = await mixerApi.get(`channels/${channelId}/relationship?user=${userId}`, "v1", false, true);

            if (userRelationshipData != null) {
                userFollowsChannel = userRelationshipData.status && userRelationshipData.status.follows != null;
            } else {
                userFollowsChannel = false;
            }
            // set cache value
            followCache.set(`${userId}:${channelNameOrId}`, userFollowsChannel);
        }

        if (!userFollowsChannel) {
            userfollowsAllChannels = false;
            break;
        }
    }
    return userfollowsAllChannels;
}
```

LastMageTV: Okay, I made this update. Only thing I changed is that if the username or ID doesn't exist, it returns false instead of continuing to check others. I put in a logger warning message if that happens, so we can track down issues if someone typos something and thinks its broken.

PR #931, opened 2020-08-22
Heist bugfix

senepa:
LayDwnClown noted an issue with the minimum team size in the Heist command.

"Hello there. having an issue with the Heist game. Whenever a viewer runs it solo it never starts the heist. the bot acts like it wasn't enough viewers to run. But i have the min set to 1, there are not role restrictions on the heist game or heist command. But if i run it as the streamer it runs fine. Or if more than 2 folks join the heist it runs fine. Am I missing anything?"
(https://discordapp.com/channels/372817064034959370/372818514312167424/746620500373143644)

When a lobby is created within the game, a comparison happens between a user set minimum size and the user list before the user requesting the heist has been added to the user list.

This patch fixes the comparison without a reorganization of the heist-command


ebiggz:
Great catch. Thanks for this!!

backend/games/builtin/heist/heist-runner.js
```diff
@@ -141,7 +141,7 @@ exports.triggerLobbyStart = (startDelayMins) => {
 
         const heistSettings = gameManager.getGameSettings("firebot-heist");
         const minTeamSize = heistSettings.settings.generalSettings.minimumUsers;
-        if (usersInHeist.length < minTeamSize) {
+        if (usersInHeist.length < minTeamSize - 1) { // user is added to usersInHeist after triggerLobbyStart is called in heist-command
```
ebiggz, line 144: Just a heads up for the future, we try to keep our line lengths shorter than 120 characters. So we typically put code comments the line above the code. 

We don't have this documented anywhere or in our ES Lint rules so I realize theres no way you would've known this.

Issue #964, opened 2020-10-26
[Feature Request] Add clear cooldown option to Cooldown Command effect


CrowbarToolsRobot:
@ebiggz via Discord:

Add clear cooldown option to Cooldown Command effect

PR #965, opened 2020-10-26
#964 Create a clear option to clear command cooldowns.

MageEnclave:
Implements #964 

This adds in a clear option for the command cooldowns effect. You can clear the global cooldown, a user specific cooldown, or both.

Changes:
- Add "action" option to clear command cooldown effects. It has "add" and "clear". 
- Add options to clear global and/or user specific cooldowns.


ebiggz (approved):
Dude, this is epic. Thank you!

PR #980, opened 2020-11-25
Confetti Guns - The most important pull request ever created.

MageEnclave:
This just implements a confetti option for the celebration effect. It will fire confetti guns from the bottom left and right of the screen for however long the user has defined.

Changes:
- Adds a confetti option to the celebration effect.

Uses this package from NPM.
https://github.com/catdad/canvas-confetti


ebiggz (approved):
this. is. epic. 🎉

PR #1058, opened 2021-04-16
Add an option to add time to the current date

CaveMobster:
This is not a requested feature, but I'd like to use it and I thought it might be useful for others too.

It adds the possibility to add units of time to the current date.


ebiggz (approved):
This is great, thanks for doing this!

Issue #1034, opened 2021-03-06
[Feature Request] Add "Stream Team" as a Twitch user role


CrowbarToolsRobot:
@ebiggz via Discord:

Add "Stream Team" as a Twitch user role

CaveMobster:
Seems doable, the channel API has this information. I'll take this one on.

PR #1065, opened 2021-04-19
Add team roles

CaveMobster:
### Description of the Change
This adds one or multiple team roles, depending on how many streams the streamer is in. When a streamer is not in a stream team, it shows an empty row in the viewers roles tab, and no team roles will show up anywhere else in the bot.

### Applicable Issues
#1034


ebiggz (approved):
Fantastic work, thank you so much!

PR #1067, opened 2021-04-20
Add a regex variable

CaveMobster:
### Description of the Change
This adds a variable to use regular expressions.

### Applicable Issues
n/a


ebiggz (changes requested):
Great idea for a $variable. 

Based on feedback I got on the $textReplace variable which also supports regex, we will want to add the ability for the user to define regex flags too.

```js
evaluator: (_, stringToEvaluate, expression, flags) => {
        const regex = RegExp(expression, flags);
```

CaveMobster:
Oh good point, thank you! Added this, and also changed the return value of `null` into a clear response message when no match is found.

CaveMobster:
Updated as discussed on Discord.

ebiggz (approved):
Excellent, thanks for hashing out the details with me on this one 🥇

PR #1070, opened 2021-04-20
V5 Manual Highlight

MageEnclave:
### Description of the Change
This adds in an option to manually trigger a "message highlight" event from the Firebot chat window. The use case that inspired this was the ability to highlight a specific message in the overlay in order to bring more attention to it.

There is a "highlight message" channel points reward that I didn't think about until this was completed. Do you think the naming of this event may be confusing at all?

### Applicable Issues
N/A

### DIFF
--- a/gui/app/services/chatMessagesService.js
+++ b/gui/app/services/chatMessagesService.js
@@ -122,6 +122,16 @@
                 }
             };
 
+            service.highlightMessage = (username, rawText) => {
+                backendCommunicator.fireEvent("firebot:highlight-message", {
+                    action: "highlightMessage",
+                    meta: {
+                        username: username,
+                        messageText: rawText
+                    }
+                });
+            };
+
             // Chat Alert Message
             service.chatAlertMessage = function(message) {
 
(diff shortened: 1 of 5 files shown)


gui/app/services/chatMessagesService.js
```diff
@@ -122,6 +122,13 @@
                 }
             };
 
+            service.highlightMessage = (rawText) => {
+                backendCommunicator.fireEvent("highlight-message", {
+                    action: "highlightMessage",
+                    messageText: rawText
+                });
+            };
```
ebiggz, line 130: We'll need to update this function to the following:
```js
service.highlightMessage = (username, rawText) => {
    backendCommunicator.fireEvent("highlight-message", {
        username: username,
        chatMessage: rawText
    });
};
```
What I changed:
- Adds `username` as the first arg so we can pass it along in the event data. You're already calling this properly in that switch statement (`chatMessagesService.highlightMessage(userName, rawText);`)
- Added `username` property to the event data object.
- Removed the `action` property as its not needed
- updated `messageText` to `chatMessage` to match what the `$chatMessage` var is expecting

ebiggz (changes requested):
Thanks for doing this!!

There's just a couple small tweaks needed for one of the functions and then this will be good to go 👍

ebiggz (approved):
Epic thank you!

Issue #1051, opened 2021-04-13
[Bug] Chat Message event is not triggered when chatting as the Streamer Account via Firebots Chat Feed


CrowbarToolsRobot:
@ebiggz via Discord:

Chat Message event is not triggered when chatting as the Streamer Account via Firebots Chat Feed

PR #1071, opened 2021-04-20
Fix message event not triggering on streamer chatting in firebot chat.

MageEnclave:
### Description of the Change
This makes it so that the chat message event will trigger when the user chats as the streamer account in Firebot's chat window.

### Applicable Issues
#1051 

### Testing
I tested by creating a !test command that had my streamer account reply with who chatted. I then typed !test into the firebot chat window and got the response that I chatted. This did not trigger the chat message event a second time, which is what I would expect.


ebiggz (approved):
Thanks for fixing one of my bugs 🙏

Issue #1064, opened 2021-04-19
[Feature Request] Add a way to easily get someones sub tier


CrowbarToolsRobot:
@ebiggz via Discord:

**Example use case:** 
Play fanfare when a tier 2 or 3 arrives in chat. 

**Possible solutions:**
 1. `$subLevel[username]` variable
 2. Adding Tier 1,2,3 and prime twitch roles
 
 I think **1** would probably be the easiest to implement and the most versatile but I am open to **2** or other ideas!

CaveMobster:
I think 2 would be the friendliest for users, kraken offers information about channels that a user is subscribed to, that includes the sub type (prime, tier 1, tier 2, tier 3). We could split the subscriber role into the different types so that the information is available in different parts of the bot like permissions and the viewer details.

ebiggz:
@CaveMobster I can definitely see **2** being more useful. Once **2** is implemented, it would be easy to implement **1** later if we ever needed. Go for it!

PR #1072, opened 2021-04-20
Add subscriber tier roles

CaveMobster:
### Description of the Change
This adds three new subscriber roles to the Twitch roles: tier 1, tier 2 and tier 3. Prime subs are not available to pull from the API, but I think Prime is mostly interesting in subscribing events, and it's already available in there.

### Applicable Issues
Add a way to easily get someones sub tier https://github.com/crowbartools/Firebot/issues/1064


ebiggz (approved):
Nice work!

PR #1073, opened 2021-04-20
Add an effect for adding and removing VIP badges

CaveMobster:
### Description of the Change
Adds an effect with which you can add or remove VIP badges, similar to mod/unmod. Useful for streamers that have the VIP badge as a channel point redemption or as a subscriber perk.

### Applicable Issues
n/a


ebiggz (approved):
You're knocking stuff out like crazy! Looks great

CaveMobster:
What can I say... I enjoy working on it a lot. 😄

Issue #1019, opened 2021-01-14
[Feature Request] Support "Event Sets" in Setups


CrowbarToolsRobot:
@ebiggz via Discord:

Support "Event Sets" in Setups

PR #1075, opened 2021-04-21
Add event sets to the setup creation

CaveMobster:
### Description of the Change
Adds event sets as a separate option in the setup creation.

### Applicable Issues
Support "Event Sets" in Setups https://github.com/crowbartools/Firebot/issues/1019


ebiggz (commented):
Thanks for jumping on this!

We will also need to update the setup import logic to account for the Event Sets. That'll need to be added to [this function](https://github.com/crowbartools/Firebot/blob/b23c9dbcf531c86588d61ba07ab448b1e36f921e/backend/import/setups/setup-importer.js#L74).

I am happy to do that, unless you'd like to handle it. Either way, let me know!

CaveMobster:
Ah yep! I'll handle it, good for the learning experience. Thank you!

CaveMobster:
Fixed! It seems I also forgot to add data to the confirmation modal after importing. 😅 I think my excitement has been getting in the way of properly checking and testing everything.

ebiggz (approved):
Looks great🥇

PR #1095, opened 2021-04-25
Make Counter Manager Available to Custom Scripts

pirak:
### Description of the Change
Currency, custom variable, game, timer, … managers are already available to custom scripts. The counterManager was missing. This PR adds it.

Adding the counter-id in a custom script as parameter is not really nice, by adding the `counterManager` only the name has to be added, the script can then query the id from there.

### Applicable Issues
n/a


ebiggz (approved):
Thanks for doing this!

Issue #1104, opened 2021-04-27
[Bug] Chat message variable doesn't work with events.


MageEnclave:
**Describe the bug**
I think something got mixed up at the end of dev for the manual highlight feature. A variable name was changed, but it wasn't updated everywhere it needed to be. Right now the highlight feature can't parse $chatMessage.

In chat-message.js line 34.

**Steps To Reproduce**
1. Create a new chat highlight event.
2. Set it up to output a chat message containing $username and $chatMessage.
3. Try to highlight a message and you'll see neither variable parses because of a $chatMessage parsing error.

**Expected behavior**
All variables should be parsed without errors.

**Screenshots**
![image](https://user-images.githubusercontent.com/17416635/116246054-3a660c80-a72f-11eb-8bdd-f605b2eb4e82.png)

## Contact Us!
Did you know that we have a help channel in discord as well? https://discord.gg/DegW792

PR #1107, opened 2021-04-27
#1104 V5 chat message variable fix

MageEnclave:
### Description of the Change
From what I can tell, events seem to use messageText where other things tend to use chatMessage. We tried updating messageText to chatMessage for the release of manual highlighting, but it caused problems.

This changes chatMessage back to messageText which should fix the $chatMessage variable for events.

### Applicable Issues
#1104 

### Testing
1. Create a new highlight message or chat message event.
2. Have it output a chat message or overlay text.
3. Put in a $chatMessage variable in the effect.
4. Fire the event and confirm chatMessage gets parsed correctly.

Tested on message highlight and chat message events.

image

### DIFF
--- a/backend/variables/builtin/chat-message.js
+++ b/backend/variables/builtin/chat-message.js
@@ -31,7 +31,6 @@ const model = {
 
         } else if (trigger.type === EffectTrigger.EVENT) {
             // if trigger is event, build chat message from chat event data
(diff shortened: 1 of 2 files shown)


ebiggz (approved):
Sorry about this - looks like it was ultimately my fault for suggesting the property name change to `chatMessage`.

Issue #1068, opened 2021-04-20
[Feature Request] Enable/disable command effect


CrowbarToolsRobot:
@ebiggz via Discord:

We need an effect to toggle the enabled state of a command.

PR #1122, opened 2021-05-01
Add an effect to toggle a command

CaveMobster:
### Description of the Change
This adds an effect that lets the user toggle a system command or a custom command through an effect.

### Applicable Issues
Enable/disable command effect https://github.com/crowbartools/Firebot/issues/1068

### Testing
Added toggle effects for both system and custom commands, checked if they updated correctly and if the UI updated too.

### Screenshots
image
image


CaveMobster:
@ebiggz This gave me a headache haha, but I ended up using the command manager instead of command access, because through command access there was no way to update the UI for system commands. It's ready for review now!

backend/effects/builtin/toggle-command.js
```diff
@@ -0,0 +1,112 @@
...
+                        $scope.effect.command = command;
+                    }
+                }
+            }
+        };
+
+        $scope.toggleOptions = {
+            disable: "Deactivate",
+            enable: "Activate"
+        };
```
ebiggz, line 76: Lets add a third "toggle" option as well.

ebiggz, line 76: Lets also add a third "toggle" option here

backend/effects/builtin/toggle-command.js
```diff
@@ -0,0 +1,112 @@
...
+            commandManager.saveSystemCommandOverride(command);
+        }
+
+        if (command.type === "custom") {
+            command.lastEditAt = moment().format();
+
+            commandManager.saveCustomCommand(command, event.trigger.metadata.username, false);
+        }
+
+        commandAccess.triggerUiRefresh();
```
ebiggz, line 106: We should be able to get the most current command/syscommand using the `commandId` and `commandType` effect properties instead of caching the entire command in the effect which could lead to stale data (see comment above).

Here's how we could accomplish that:
```js
const { commandId, commandType, toggleType } = effect;

if (commandType === "system") {
    const systemCommand = commandManager
        .getAllSystemCommandDefinitions().find(c => c.id === commandId);

    if (systemCommand == null) {
        // command doesnt exist anymore
        return true;
    }

    systemCommand.active = toggleType === "toggle" ? !systemCommand.active : toggleType === "enable";

    commandManager.saveSystemCommandOverride(systemCommand);

    frontendCommunicator.send("systemCommandsUpdated");
} else if (commandType === "custom") {
    const customCommand = commandManager.getCustomCommandById(commandId);

    if (customCommand == null) {
        // command doesnt exist anymore
        return true;
    }

    customCommand.active = toggleType === "toggle" ? !customCommand.active : toggleType === "enable";

    commandManager.saveCustomCommand(customCommand, "System", false);
}
```

ebiggz (changes requested):
Thank you for working on this! The command stuff is indeed a HUGE headache. It has some of our oldest and most archaic code so I commend you for wading through it. 

I have left a couple suggests, mainly revolving around how we handle the command data. Please let me know if you have any concerns or questions with these!

backend/effects/builtin/toggle-command.js
```diff
@@ -0,0 +1,112 @@
...
+
+        $scope.updateSelectedCommand = commands => {
+            if ($scope.effect.commandId) {
+                for (const command of commands) {
+                    if (command.id === $scope.effect.commandId) {
+                        $scope.effect.command = command;
+                    }
+                }
+            }
+        };
```
ebiggz, line 71: We'll want to avoid storing the entire command object inside the effect data as it will eventually lead to stale data that can cause unexpected behavior.

Example of potential issue: User creates a Toggle Command effect, then later they edit the command to change its cooldown. Now when that Toggle Command effect runs, it will override all data for that command including replacing the cooldown with the old value.

Instead we should be able to rely on the already present `effect.commandId` and `effect.commandType` to grab current command data at effect run time. I detail this below.

backend/effects/builtin/toggle-command.js
CaveMobster: Ahhh this is perfect! Thank you so much for explaining all this, and for showing me the correct solution. It's really helpful and educational. I updated the PR according to all of the feedback, including the toggle state. 😄

PR #1128, opened 2021-05-04
Add a button to the system commands list that toggles the active state of a command

CaveMobster:
### Description of the Change
Adds a button to the system commands list that toggles the active state of a command, so that you don't have to open the modal first.

### Applicable Issues
n/a

### Testing
Clicked the button for both state, checked if in the edit modal if "Is Active" was in the right state.

### Screenshots
image


ebiggz (approved):
@CaveMobster Thanks for doing this!

It's worth mentioning that you can also right-click the system commands to quickly edit or toggle them (you can also do this for custom commands, events, and timers). Seems we need to do something to make this feature more easily discovered by users 🤔 
image

Issue #1130, opened 2021-05-05
[Bug] Viewer's Roles condition is not working for Stream Team Roles


CrowbarToolsRobot:
@ebiggz via Discord:

Looks like we need to await a promise that is getting returned by `teamRolesManager.getAllTeamRolesForViewer(username)` https://github.com/crowbartools/Firebot/blob/master/backend/effects/builtin/conditional-effects/conditions/builtin/viewer-roles.js#L69

PR #1131, opened 2021-05-05
Fixes stream teams not working in the Viewers Roles condition + general clean up

ebiggz:
### Description of the Change

The stream team roles were not working in the Viewers Roles condition because `teamRolesManager.getAllTeamRolesForViewer` returns a Promise that we were not awaiting on. 

While I was at it I also did some general clean up of really old code + added some JSDocs.

@CaveMobster I am requesting a review from you as you worked in this recently and are the most familiar, so let me know if I forgot anything. I am also wondering if you would be willing pull down this branch and briefly test the Viewers Roles condition to confirm the fix. I am not in any stream teams so I cannot do so myself :( 

### Applicable Issues

#1130

### DIFF
--- a/backend/database/currencyDatabase.js
+++ b/backend/database/currencyDatabase.js
@@ -151,12 +151,19 @@ function addCurrencyToUserGroupOnlineUsers(roleIds = [], currencyId, value, igno
 
         const onlineUsers = await userDatabase.getOnlineUsers();
 
+        /** @type{ Record<string, Array<{ id: string; name: string }>> } */
+        const teamRoles = {};
+        for (const user of onlineUsers) {
+            teamRoles[user.username] = await teamRolesManager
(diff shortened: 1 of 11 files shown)


CaveMobster:
Looks great! I forgot the await everywhere I used that function, so might be nice to fix that while we're at it. 😅 
`backend\database\currencyDatabase.js`
`backend\events\filters\builtin\viewerRoles.js`
`backend\games\builtin\heist\heist-command.js`
`backend\games\builtin\slots\spin-command.js` 
`backend\games\builtin\trivia\trivia-command.js`
`backend\restrictions\builtin\permissions.js`

ebiggz:
Ah thanks for pointing that out! I think I got all instances updated 🎉

CaveMobster (approved):
Great improvements!

PR #1137, opened 2021-05-07
Add a variable for total bits cheered in the channel

CaveMobster:
### Description of the Change
This adds a variable to display the total amount of bits cheered in the channel by the associated user.

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1127

### Testing
I tested with the test function, unfortunately no way to do live testing.

Issue #1032, opened 2021-03-04
[Feature Request] New Event Type: 'Whisper'


CrowbarToolsRobot:
@heyaapl via Discord (on behalf of @MrCanerican in Discord):

New Event Type: 'Whisper' which, when a Streamer Account is whispered, Effects may be triggered.

heyaapl:
Duplicate of #1033 - Closing

ebiggz:
Reopening this in lieu of #1033 . I think a new event rather than a filter would actually be better. Otherwise chat events with no filters would suddenly start to get triggered by whispers and I could foresee that being bad.

PR #1138, opened 2021-05-07
Add an event to catch whispers

CaveMobster:
**Important:** Please look at https://github.com/crowbartools/Firebot/pull/1137 (total bits) first, as there are going to be conflicts.

### Description of the Change
This adds an event to catch whispers sent to the streamer.

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1032

### Testing
I made an event with a chat alert effect that includes both username and the message, then whispered my streamer account from my bot account.


ebiggz (dismissed):
Excellent!

Issue #1167, opened 2021-05-11
[Feature Request] Url chat moderation


CrowbarToolsRobot:
CaveMobster via Discord:

Url moderation would be useful. Messages containing urls could be automatically deleted, and allowed based on viewer role or watch time. A permit command as a system command when url moderations is enabled, could allow urls for a period of time.

PR #1198, opened 2021-05-24
Add a url moderation feature to the moderation tools

CaveMobster:
### Description of the Change
This adds a url moderation feature to the moderation tools. When view time is enabled, it will check for both exempt roles and view time, otherwise it will only check for exempt roles. 

It also automatically adds a !permit system command with which viewers can be temporarily be permitted to post urls, the default duration (30 seconds) of that permit can be adjusted by the user.

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1167

### Testing
The following cases were tested:
- Viewer time disabled with and without message and variables
- Viewer time enabled with and without message and variables
- Exempt roles active with and without viewer time enabled
- Permit command with and without message and variables
- Permit command with different permit durations
- Permit command with multiple users

### Screenshots
image
image

### DIFF
--- a/backend/chat/chat-listeners/listeners/chat-message-streamer-listener.js
+++ b/backend/chat/chat-listeners/listeners/chat-message-streamer-listener.js
@@ -14,7 +14,7 @@ module.exports = {
         const timerManager = require("../../../timers/timer-manager");
 
         //Send to chat moderation service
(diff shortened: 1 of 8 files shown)


ebiggz (approved):
This is badass! Thanks for doing this

Issue #755, opened 2020-03-11
[Feature Request] Allow editing quote date to null


CrowbarToolsRobot:
@TheLastMage via Discord:

Allow users to edit the date on a quote.

CaveMobster:
This is already in the bot, right?

ebiggz:
The original title was confusing. what the user wanted was the ability to set a quote date to empty

PR #1202, opened 2021-05-27
Allow users to save empty dates on quotes

jstnrp:
### Description of the Change

Allow users to save empty dates on quotes, sets empty date to [No Date]

### Applicable Issues

#755


ebiggz (approved):
This looks great, thanks for doing it!

Issue #1203, opened 2021-05-27
[Feature Request] Show channel reward cost in table


CrowbarToolsRobot:
ebiggz via Discord:

Setari suggested it might be good to show the channel reward costs at the top level table!

PR #1204, opened 2021-05-27
Add channel point cost to the main table

jstnrp:
### Description of the Change

Add channel point cost to the main table

### Applicable Issues

#1203


ebiggz (approved):
Perf

Issue #1195, opened 2021-05-21
[Tech Debt] Refactor all pages to use <firebot-item-table>


CrowbarToolsRobot:
@ebiggz via Discord:

Refactor all pages to use `<firebot-item-table>`

It will solve:
- Duplicate code and the maintainability issues that brings
- Overflow dropdown menu issues
- Add sorting to everything that uses it

zunderscore:
Released in v5.62.0

Issue #1084, opened 2021-04-22
[Feature Request] Group preset effect lists


CaveMobster:
**Describe the solution you'd like**
Currently the preset effect lists are one big list, but that can get confusing really fast if you use a lot of them for a variety of reasons. For example, I have an end credits system that uses multiple preset effect lists, but I also have one for sending out a go-live message, and one to clear sound effects. Having groups would help to keep things organized. Event sets would be a good inspiration for the UI.

ebiggz:
This is a good idea!!

However I think we should start with implementing "Sort Tags" (used in Commands and Events) first and see if that is sufficient. "Event Sets" are special in that there is app functionality tied to them: they allow the user to quickly enable or disable groups of events which is useful for having different groups of events for different games, holidays, etc. So Sets are different from UI organization. Sort tags are intended to be the mechanism for sorting/organization within the UI and its why Events have both Sort Tags and Sets.

CaveMobster:
Oh that's a great point! So if I understand this correctly, with the sort tags it would be more like how the commands are sorted right?

ebiggz:
@CaveMobster Yep, exactly like that. Both Events and Commands use Sort Tags:

![image](https://user-images.githubusercontent.com/3515161/115939302-53d33400-a463-11eb-9a6b-cc5eb4765411.png)

![image](https://user-images.githubusercontent.com/3515161/115939342-77967a00-a463-11eb-8ba6-04ebf2d65512.png)

CaveMobster:
@ebiggz I'd like to start working on this soon, so I did some looking into the code. Do you think it's a good idea to convert the Preset Effect Lists view to the Firebot item table that the Channel Rewards use before adding sort tags? And if so, is it okay if I do that, or would that mess with your planning?

ebiggz:
@CaveMobster Yep go for it! I created the `<firebot-item-table>` element to be as reusable as possible so we can use it for all pages that share the same UI (something that should of happened long ago).

PR #1225, opened 2021-06-11
Update the preset effect lists with the firebot item table

CaveMobster:
### Description of the Change
This changes the preset effect lists view to make use of the firebot item table, and adds sort tags, manual sorting and duplication of preset effect lists.

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1084 and https://github.com/crowbartools/Firebot/issues/1195

### Testing
In the effects tab, I tested:
- name changes
- effect changes
- duplicating items
- reordering items
- deleting items
- searching for items
- creating and deleting sort tags
- adding and removing sort tags to the items
  
In the effects window when adding an effect list from another place such as commands, I tested if it fetched the correct list, and if it updated and returned the correct list when editing it from there.

### Screenshots
image


ebiggz (approved):
Fantastic work! Thank you so much for doing this!

Issue #1241, opened 2021-07-07
[Bug] Multimonth gift sub shows gifter name instead of gift receiver name


CrowbarToolsRobot:
CaveMobster via Discord:

This seems to go wrong in the twitch-pubsub-client package, it returns the gifter's username if the context was not `subgift`. And that's correct, because the context for this message is probably `resubgift`.

PR #1242, opened 2021-07-07
Add a workaround to return the correct username for multi month gift subs

CaveMobster:
### Description of the Change
This workaround takes care of two things:
1. It checks for the right context in the pubsub listener, since `isGift` only checks for `subgift`, and not for `resubgift`, `anonsubgift` and `anonresubgift`.
2. Since the package's `userDisplayName` also only checks for the `subgift` context, it now tries to access `recipient_display_name` straight from the data, and falls back to `userDisplayName` if it can't find it.

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1241


CaveMobster:
Sorry, closed it because I realized this is in fact

ebiggz (approved):

PR #1277, opened 2021-08-14
Add a test button for the preset effect lists

CaveMobster:
### Description of the Change
This adds a test button to the preset effect lists tab.

### Applicable Issues
n/a

### Testing
Tested multiple effects with the button.


ebiggz:
Glad the firebot item table is coming in handy 🚀

PR #1281, opened 2021-08-15
feat: added $encodeForHtml for safe variable use in HTML templates

FineWolf:
### Description of the Change
Added the built-in variable $encodeForHtml which properly escapes HTML entities.
This allows the safe use of $variables within HTML templates for the HTML effect.

### Testing

- Built local version
- Tested the variable within an HTML template
- Tested injection scenarios to make sure they are not possible (cheerMessage: `hello <strong>world</strong>`)

#### New built-in variable in the `$vars` picker:
image

#### Extended warning at the bottom of the HTML effect
image


FineWolf:
I've changed the base branch for my commit to `v5` as per the other pull requests in this repo.

ebiggz (dismissed):
Man this will really come in handy. Great idea. Thanks for taking the time to implement it!

ebiggz (approved):

Issue #1288, opened 2021-08-16
[Bug] After executing an "Effect List" effect that is running a preset, event variables are no longer properly parsed.


FineWolf:
**Describe the bug**
When executing an "Effect List" that is set to run a preset list, the following effects in the chain are no longer able to access event variables.

**Steps To Reproduce**
1. Create a new event (Subscription for example)
2. Add an "Effect List" as the first effect in the chain, configure it to use a preset.
3. Add a effect following the effect list that uses an event variable (example: TTS using $subMonths)

`$subMonths` will fail to evaluate.

**Expected behavior**
Event variables evaluate properly down the chain.

**Screenshots**

**Firebot Version:** v5.40.1

**Computer OS:** Windows 10

**Note:** Will self implement fix

PR #1289, opened 2021-08-16
fix: trigger being mutated in effectGroup causing issues for effects …

FineWolf:
…down the chain

### Description of the Change
The "Run Effect List" (effectGroup) effect was incorrectly causing a mutation which was preventing subsequent effects in the chain from being able to evaluate event variables.

### Applicable Issues
#1288  (self-reported)

### Testing
Simulated events to ensure that effects after a preset group are able to resolve variables.


ebiggz (approved):
Good catch, I know this particular bug was plaguing a number of people. Thank you!

Issue #1280, opened 2021-08-15
[Feature Request] Add option to edit tags in the edit stream info modal


CaveMobster:
**Describe the solution you'd like**
Show the tags in the dashboard stream info edit modal, as well as add the option to edit them.

PR #1315, opened 2021-08-18
#1280 stream info tags

CaveMobster:
### Description of the Change
This adds an option in the stream info modal to add and remove tags for the channel. 

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1280

### Testing
- Added tags to the max of 5, and removing them to see if the UI behaves correctly
- Tried adding double tags to verify that they don't get added
- Saved an empty tag list and a populated tag list
- Clicked the Add Tag button before the tags were loaded to make sure it doesn't open the select modal
- Checked accessibility (as much as possible with the current components)

### Screenshots
5 tags (the maximum that Twitch allows)
image

Less than 5 tags
image

Tag selecter
image

### DIFF
--- a/backend/twitch-api/resource/stream-tags.js
+++ b/backend/twitch-api/resource/stream-tags.js
@@ -4,6 +4,7 @@ const twitchApi = require("../client");
 const { TwitchAPICallType } = require('twitch/lib');
 const accountAccess = require("../../common/account-access");
 const logger = require('../../logwrapper');
+const { snakeKeys } = require('js-convert-case');
 
 /**
  * @typedef TwitchStreamTag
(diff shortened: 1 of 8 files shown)


backend/twitch-api/resource/stream-tags.js
```diff
@@ -0,0 +1,120 @@
...
+const logger = require('../../logwrapper');
+const { snakeKeys } = require('js-convert-case');
+
+/**
+ * @typedef TwitchStreamTag
+ * @property {string} id - The ID of this tag
+ * @property {boolean} isAuto - Whether this tag is automatically applied to the channel
+ * @property {string} name - A dictionary that contains the localized names of the tag
+ * @property {string} description - A dictionary that contains the localized descriptions of the tag
+ */
```
ebiggz, line 15: Thank you for using JSDocs. Makes code so much easier to work with. 🙏

ebiggz (approved):

ebiggz (approved):

PR #1326, opened 2021-08-24
Spam raid protection command

CaveMobster:
### Description of the Change
This PR adds a system command that combats hate or spam raids. The following options are available:
- Followers only mode
- Subscribers only mode
- Emote only mode
- Slow mode
- Clear chat
- Ban raiders
- Block raiders

When ban or block raiders is activated, the raid message will be filtered out and all users in the last 50 messages that used the command will be blocked and/or banned. Every user that uses the message after the command is activated will also be blocked and/or banned.

### Applicable Issues
n/a

### Testing
I tested all options separately as well as combined. I had multiple people helping to test the ban and block options and it works as intended. 

### Screenshots
image


backend/chat/moderation/raid-message-checker.js
```diff
@@ -0,0 +1,90 @@
...
+    let rawMessages = messageCache.map(message => message.rawText);
+    let raidMessages = rawMessages.reduce(function (allMessages, message) {
+        if (message in allMessages) {
+            allMessages[message]++;
+        } else {
+            allMessages[message] = 1;
+        }
+
+        return allMessages;
+    }, {});
```
ebiggz, line 57: LOVE the use of `.reduce`. Couple relatively small stylistic things I would tweak however:
1. Use a fat arrow function to be consistent with everything else in `getRaidMessage` 
```js
rawMessages.reduce((allMessages, message) => {
```

2. Instead of using `if(message in allMessages)`, use `if(allMessages[message] != null)`. 
The `in` operator only checks for the existence of a property, it doesn't check the value for that property. Showcase of issue:
```js  
const foo = { bar: undefined };
console.log("bar" in foo); // outputs: true
```
> Note it returns true even though the property is undefined which wouldn't be good if we were trying to increment a number by 1. This shortcoming wouldn't be a problem where you are currently using it, but it's just a best practice to avoid using that all together so you never have to work about its gotcha's. 


3. Speaking of gotchas, it's typically best to use `+= 1` instead of `++` to increment a number by 1. This is because ++ has really weird behaviors associated to it that can make unexpected results happen if you aren't careful.

Gotcha showcase:
```js
function incrementTest() {
    let num = 1;
    return num++;
}
console.log(incrementTest()); // Output: 1
```
> Note it returned 1 instead of 2. This is because ++ doesn't increment until AFTER the function returns. Weird right? I honestly don't know why it behaves this way. In any case, `+= 1` doesn't have this drawback and has the added benefit of being less ambiguous as well.

ebiggz (changes requested):
Looks fantastic! Just a couple small stylistic suggestions but otherwise looks good to go functionally! 

Side note: don't feel obligated to do this for now but going forward, try to make sure to always use `const` unless a variable need to be able to be reassigned, in which case you can use `let`. I know a lot of our old code uses `let` so it's not a big deal. Just a good best practice going forward. :)

backend/chat/moderation/raid-message-checker.js
CaveMobster: Woahh, good points!! Updated the PR to use all of your suggestions, and I also fixed the const/let thing. Thank you!!

ebiggz (approved):

PR #1351, opened 2021-09-03
Moderation worker refactor

CaveMobster:
### Description of the Change
This adds emote limit and url moderation to the moderation service. This also fixes a bug where banned regexes couldn't be removed.

### Testing
Tested if all mod tools work as intended, with several combinations and settings.


SReject (changes requested):
In the [moderation service](https://github.com/crowbartools/Firebot/blob/moderation-worker-refactor/backend/chat/moderation/moderation-service.js#L12) `bannedWords` is initialized as an array, which does not have a `.has()` function.

Assuming you went off my example, if you plan to use a set, bannedWords should be initialized as a `Set`

SReject (approved):

PR #1355, opened 2021-09-04
Spam raid command options

CaveMobster:
### Description of the Change
This moves the spam raid protection functionality to the moderation worker, as well as adds options in the moderation tab to change the cache limit and the amount of characters to cut off at the start of a message.

### Testing
Tested if the protection functionality as well as all other features still work, also tested if the setting changes work with different values.


ebiggz (changes requested):
Looks great although there's a merge conflict.

ebiggz (approved):

Issue #1333, opened 2021-08-25
Switch from Kraken to Helix API 


CaveMobster:
Kraken will be deprecated in February 2022: https://dev.twitch.tv/docs/v5
We use Kraken in several places, so we need to find a way to fully switch to Helix.

PR #1359, opened 2021-09-07
#1333 kraken to helix

CaveMobster:
### Description of the Change
This PR does the following:
- Updates the twitch API package
- Changes all Kraken API calls to Helix calls 
- Cleans up Helix calls like paginated block lists and stream tags
- Updates several Helix calls that were not able to use the package methods yet
- Moves and reworks the chat role functionality to a separate chat roles manager, since Helix doesn't offer a way to get chat roles
- Moves all team roles logic to the team roles manager, so that all team role information will be accessed through there
- Fixes a bug in the active viewer list effect
- Removes the teams api, the channel progression restriction and the channel access

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1333

### Testing
Tested all functionality that was touched and updated, and checked for backwards compatibility.


backend/roles/chat-roles-manager.js
```diff
@@ -0,0 +1,133 @@
...
+
+        if (moderators) {
+            for (const moderator of moderators) {
+                if (moderator.userId === userId) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
```
ebiggz, line 27: we can shorten this to:
```js
return moderators ? moderators.some(m => m.userId === userId) : false;
```

backend/roles/chat-roles-manager.js
```diff
@@ -0,0 +1,133 @@
...
+
+        if (vips) {
+            for (const vip of vips) {
+                if (vip === username) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
```
ebiggz, line 47: We can shorten this similarly to the mods suggestion:
```js
return vips ? vips.includes(username) : false;
```

backend/roles/chat-roles-manager.js
```diff
@@ -0,0 +1,133 @@
...
+            roles.push("mod");
+        }
+
+        if (userIsVip) {
+            roles.push("vip");
+        }
+
+        if (subRoles) {
+            roles = roles.concat(subRoles);
+        }
```
ebiggz, line 123: subRoles will never be null (it returns array or empty array) so this check isn't needed.

backend/roles/team-roles-manager.js
```diff
@@ -2,45 +2,62 @@
...
-                id: team.id,
+                id: parseInt(team.id), // For compatibility (Kraken API returned an int)
                 name: team.displayName
             };
         });
-}
-
-async function getAllTeamRolesForViewer(username) {
-    const roles = await twitchApi.teams.getMatchingTeamsByName(username);
+};
```
ebiggz, line 22: Great thinking on making sure this remains compatible. 

Super minor, but we can strip away some unneeded syntax like so
```js
const mapRoles = (teams) => teams.map(t => ({
  id: parseInt(t.id),
  name: t.displayName
})
```

backend/roles/team-roles-manager.js
```diff
@@ -2,45 +2,62 @@
...
-}
+/** @returns {Promise.<MappedTeamRole[]>} */
+const getAllTeamRolesForViewer = async (username) => {
+    const client = twitchApi.getClient();
+    const user = await client.helix.users.getUserByName(username);
+    const streamerTeams = await client.helix.teams.getTeamsForBroadcaster(accountAccess.getAccounts().streamer.userId);
 
-async function getTeamRoles() {
-    const teams = await twitchApi.teams.getStreamerTeams();
+    if (streamerTeams == null) return null;
```
ebiggz, line 30: Might make sense to just return an empty array here instead of null so any code that calls this function doesn't have to worry about null checks. That's typically my rule of thumb with functions that returns arrays.

backend/roles/team-roles-manager.js
```diff
@@ -2,45 +2,62 @@
...
-    if (teams == null) {
-        return null;
+    const teams = [];
+    for (const team of streamerTeams) {
+        const relations = await team.getUserRelations();
+        for (const relation of relations) {
+            if (relation.id === user.id) {
+                teams.push(team);
+            }
+        }
```
ebiggz, line 39: The outer `for` loop is unfortunately necessary because of the `await` call, but we could remove this nested one by instead doing:
```js
if(relations.some(r => r.id === user.id)) {
   teams.push(team);
}
```

backend/twitch-api/resource/stream-tags.js
```diff
@@ -14,108 +12,55 @@ const { snakeKeys } = require('js-convert-case');
...
-            return null;
+        if (response) {
+            /**@type {TwitchStreamTag[]} */
+            return response.filter(tag => !tag.isAuto).map(tag => mapTwitchTag(tag));
         }
-
-        return response;
     } catch (error) {
         logger.error("Failed to get all stream tags", error);
         return null;
```
ebiggz, line 35: I know this is a previously exiting function. Just wanted to mention the returning an empty array instead of null thing again. Not sure how it would affect what is calling this though.

backend/twitch-api/resource/stream-tags.js
```diff
@@ -14,108 +12,55 @@ const { snakeKeys } = require('js-convert-case');
...
-                "broadcaster_id": accountAccess.getAccounts().streamer.userId
-            },
-            body: snakeKeys(tagIds)
-        });
-        return true;
+        const client = twitchApi.getClient();
+        await client.helix.streams.replaceStreamTags(accountAccess.getAccounts().streamer.userId, tagIds);
     } catch (error) {
         logger.error("Failed to update channel stream tags", error);
         return false;
```
ebiggz, line 60: Returns false in the catch but the `return true` was removed from the try. Is the return boolean even used? If not, we can just remove the `return false` from the catch. If yes, we'll want to add the `return true` back to try.

backend/common/user-access.js
```diff
@@ -84,17 +82,17 @@ async function getUserDetails(userId) {
 
     let isBlocked = false;
     try {
-        const blocklist = await twitchApi.users.getAllBlockedUsersPaginated(streamerData.userId);
+        const blocklist = await twitchApi.getClient().helix.users.getBlocksPaginated(streamerData.userId).getAll();
         if (blocklist) {
-            isBlocked = blocklist.find(id => id === twitchUser.id) != null;
+            isBlocked = blocklist.find(user => user.userId === twitchUser.id) != null;
```
ebiggz, line 87: since we just want a boolean, instead of `.find` we can use `.some`:
```js
isBlocked = blocklist.some(u => u.userId === twitchUser.id);
```

ebiggz (changes requested):

backend/roles/chat-roles-manager.js
CaveMobster: Fixed.

backend/roles/chat-roles-manager.js
CaveMobster: Fixed.

backend/roles/chat-roles-manager.js
CaveMobster: Fixed.

backend/roles/team-roles-manager.js
CaveMobster: Fixed.

backend/roles/team-roles-manager.js
CaveMobster: Fixed.

backend/roles/team-roles-manager.js
CaveMobster: Fixed.

backend/twitch-api/resource/stream-tags.js
CaveMobster: Fixed.

backend/twitch-api/resource/stream-tags.js
CaveMobster: Fixed.

backend/common/user-access.js
CaveMobster: Fixed.

Issue #1350, opened 2021-09-03
[Bug] Connection disconnected unexpectedly


KaiShu1:
**Describe the bug**
Executing "npm start" on terminal, after streamer account login, after switching the connection, it will automatically disconnect in about a minute
  1
**Steps To Reproduce**
1. Login my twitch streamer account
2. Click to toggle connections
3. Wait minutes, it auto disconnected

**Expected behavior**
Always connected

**Screenshots**
![image](https://user-images.githubusercontent.com/44948284/131977706-02f04751-e2a8-4f50-b113-83a0d088d854.png)

**Firebot Version:**
5.40.1

**Computer OS:**
macos big sur 11.2.3

## Contact Us!
Did you know that we have a help channel in discord as well? https://discord.gg/DegW792

KaiShu1:
It doesn't look like things are getting better for me  by updating twitch-pubsub-client.
![image](https://user-images.githubusercontent.com/44948284/132078779-df70b1e7-5d62-47ba-8c48-b07d124363af.png)
![image](https://user-images.githubusercontent.com/44948284/132078794-b177fca9-3967-4d18-ae11-404605ed4a42.png)

CaveMobster:
I don't actually know if it's the pubsub client that's causing issues, or the chat client. I updated the chat client in this PR: https://github.com/crowbartools/Firebot/pull/1357 So hopefully that should help when it's merged.

KaiShu1:
> I don't actually know if it's the pubsub client that's causing issues, or the chat client. I updated the chat client in this PR: #1357 So hopefully that should help when it's merged.

OK, got it.

PR #1411, opened 2021-10-11
Fix a bug that caused the chat to fail to reconnect

CaveMobster:
### Description of the Change
This fixes a bug that caused the chat to disconnect but not reconnect. When the client is not disconnected manually it tries to automatically reconnect, but since we disconnected the client after that but didn't remove the listeners, it got stuck. Disconnecting is not needed, so I removed that line.

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1350

### Testing
Tested this by disconnecting my internet (causing error 1006) and reconnecting.

### DIFF
--- a/backend/chat/twitch-chat.js
+++ b/backend/chat/twitch-chat.js
@@ -95,10 +95,7 @@ class TwitchChat extends EventEmitter {
             });
 
             this._streamerChatClient.onDisconnect((manual, reason) => {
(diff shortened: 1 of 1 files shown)


KaiShu1:
> @KaiShu1 The `twitch-chat-client` package takes care of that, it's not within our code.

OK, got it, tks.

PR #1468, opened 2021-11-24
Trigger custom variable expire event on restart

Roger:
### Description of the Change

Trigger the event, when Firebot persist custom variables is enabled, even if Firebot was not running when the expiration happened (if not running, it will trigger the event on the next start).

### Testing

Tested the expiration of the custom variable both with Firebot running and with Firebot closed when the expiration should happen


backend/common/custom-variable-manager.js
```diff
@@ -7,13 +7,15 @@ const NodeCache = require("node-cache");
 const cache = new NodeCache({ stdTTL: 0, checkperiod: 5 });
 exports._cache = cache;
 
-cache.on("expired", function(key, value) {
+function onCustomVariableExpire(key, value) {
```
SReject, line 10: use fat arrow functions for methods that are not associated with a `this`:

```js
const onCustomVariableExpire = (key, value) => {
```

SReject (changes requested):

SReject:
Pull requests should fork from v5 and then merge against v5 :

ebiggz:
Thanks for the contribution!

Roger:
Hi! is there any guidelines for contributions? should I then rebase the PR using v5 branch instead of master?

ebiggz:
@Roger We don't have contributor guidelines at this time, unfortunately. Sorry about that! We discussed getting some defined here soon. 

But yes, rebasing to v5 would be great. We currently have it such that `master` represents the latest stable release and `v5` is our active development branch. When we are ready for a release, we merge to `master` and our CICD makes a build for us.

Roger:
Sorry I pushed an stale v5 branch from my fork, I just re-pushed updating from upstream and thanks for changing the base branch to v5!

ebiggz (approved):

ebiggz:
Looks great. I'll get this merged so it can be in the next stable release. Thanks again for contributing!

Issue #1470, opened 2021-11-24
[Bug] is not possible to use defaultValue in $customVariable


Roger:
**Describe the bug**
If `propertyPath` is not used, is not possible to use the defaultValue in $customVariable, I tough that `null` should be used as second argument if `propertyPath` was not needed, but looks like [here](https://github.com/crowbartools/Firebot/blob/master/backend/common/custom-variable-manager.js#L122) is trying to compare it as the type `null` but the argument is passed as an string `"null"`

**Steps To Reproduce**
Use anywhere something like this `$customVariable[name, null, "my default value"]`

**Expected behavior**
The custom variable replaced with "my default value" if is not set

**Firebot Version:**
v5.46.6

**Computer OS:**
Archlinux

## Contact Us!
Did you know that we have a help channel in discord as well? https://discord.gg/DegW792

Roger:
I can submit a PR for this, but I don't know if `propertyPath == null || propertyPath == "null"` is the right way to handle it, I saw that Firebot is doing those kind of comparisons already

ebiggz:
@Roger Good catch. I can't think of a markedly better approach over the `propertyPath == null || propertyPath === "null"` check. Our variable parser always returns arguments as strings so that was an oversight on my part.

Roger:
Thanks for the feedback! just created the PR

PR #1471, opened 2021-11-25
Fix using defaultValue in $customVariable

Roger:
### Description of the Change

Fix using `defaultValue` in `$customVariable`

### Applicable Issues

#1470

### Testing

Tested locally with both `$customVariable[name]` and `$customVariable[name, null, "default value"]`


ebiggz (approved):
Epic 👍

Issue #1493, opened 2021-12-02
[Feature Request] Add Day of Week options to Time Restriction


CrowbarToolsRobot:
ebiggz via Discord:

I know we initially decided to not support time frames longer than a day however a suggestion was made to support specific days of the week as some people have set streams on certain days(ie d&d on weekends). 

We should be able to add a "mode" selector to the restriction: 
- "Time Of Day"
- "Day Of Week"

"Time Of Day" is current functionality while "Day Of Week" would just show checkboxes for Monday through Sunday.

Just would need to keep backwards compatibility in mind. If the new "mode" property is not set, we must assume "Time Of Day" to retain current functionality.

ebiggz:
@Firebottle Do you feel like this enhancement is worth while and if so would you like to take it on?

Firebottle:
I can definitely add that in! This is a really good way to handle it.

Halefa:
Not sure whether I should add this as a comment here or as a new feature request: It would be awesome if the time restrictions (time, day of week) could be set for timers as well!

CaveMobster:
@Halefa I'm gonna make a new feature request for that, since it's a bit more work to implement restrictions for other triggers than first meets the eye. 😄

PR #1536, opened 2022-01-09
#1493 - Add a day selection option to the time range restriction.

Firebottle:
### Description of the Change
This updates the time restriction to allow users to select specific days in addition to a time.

### Applicable Issues
#1493 

### Testing
- Test day restriction by selecting current day.
- Test day restriction by selecting days other than the current day.
- Test day restriction by doing both of the above, but also selecting multiple days at once.
- Test time restriction by selecting a current time range.
- Test time restriction by selecting a time range outside of the current time.
- Test time restriction by selecting a time range that spans past midnight.
- Test a combination of both by applying two time restrictions to the same command. One for time and one for days.

### Screenshots
image
image

### Notes
I'm pretty sure this should be backwards compatible. It defaults to the time mode. However, it might be worth a close look at the code with that in mind.

### DIFF
--- a/backend/restrictions/builtin/time-range.js
+++ b/backend/restrictions/builtin/time-range.js
@@ -5,23 +5,86 @@ const moment = require("moment");
 const model = {
     definition: {
         id: "firebot:timeRange",
-        name: "Time Range",
-        description: "Restrict usage to a specific local time.",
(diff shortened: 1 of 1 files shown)


ebiggz (approved):
Looks great!!

PR #1537, opened 2022-01-10
Remove whisper send functionality

CaveMobster:
### Description of the Change
Whispers don't work anymore and it's causing confusion, so this will remove the functionality. 
**Note**: The change in the Command Manager is a bug fix, since by assigning the override options to the definition options, options removed from the definition would be re-assigned.

### Testing
Tested compatibility and the effects and commands that had whisper options.

### DIFF
--- a/backend/chat/commands/CommandManager.js
+++ b/backend/chat/commands/CommandManager.js
@@ -88,8 +88,6 @@ class CommandManager extends EventEmitter {
             let override = this._sysCommandOverrides[c.definition.id];
             if (override != null) {
                 if (c.definition.options) {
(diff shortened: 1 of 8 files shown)


ebiggz (changes requested):
We cannot remove this feature. I know of several people (Partners) who have gotten their Bot accounts setup as verified bots with twitch so they can use Whispers. Removing this feature would be detrimental for them. Instead of removing, we should hide the feature with an option to turn it on in settings with a description detailing the requirements.

CaveMobster:
Ohh my bad, I should have researched this better. Closing this then, and will implement the better solution later.

Issue #1029, opened 2021-02-03
[Feature Request]  Support $donationAmountFormatted for StreamElements


CrowbarToolsRobot:
@heyaapl via Discord:

$donationAmountFormatted Doesn't Work for Streamelements Donations

ebiggz:
Streamlabs and Tipeeestream both include a "formatted dono amount" in their event payload, which includes the currency sign for whatever the donator used. Unfortunately Streamelements does not include a "formatted dono amount" in its event payload, which is why the variable isn't enabled for its dono event. 

For that reason, I am going to change this bug to a feature request. I'll need to see if streamelements includes other data, such as the currency used, in the event payload so that we might be able to manually build our own formatted amount.

Runesun:
Would it make sense to break it out into two features?

1. Look into Stream Elements to see if there is a way to pull currency and formatting?
2. A way to format any number to a certain number of decimals (I don't think I know of a way to do this in Firebot currently, but maybe it exists). In lieu of option 1, I'd probably just hardcode "$" and make it a forced two decimals via generic number formatting function. It would look better than the several decimal places number that shows up now. I feel number formatting in general could probably have other uses too.

PR #1559, opened 2022-01-30
Fix the event listeners for StreamElements and add a formatted amount

CaveMobster:
### Description of the Change
It seems the websocket data has changed for actual events (so not the test data): https://dev.streamelements.com/docs/kappa/ZG9jOjYxMzcxNTY-connecting-via-websocket-using-o-auth2

I updated the listener to get the proper data, and I also added a formatter for the amount. The documentations says that `amount` is an Int32, so apparently the amount gets rounded? 

### Applicable Issues
#1029 

### Testing
I tested the follow event, the donation event is not testable unfortunately since the test functionality provides different data.

### DIFF
--- a/backend/integrations/builtin/streamelements/streamelements-event-handler.js
+++ b/backend/integrations/builtin/streamelements/streamelements-event-handler.js
@@ -55,11 +55,41 @@ exports.registerEvents = () => {
     eventManager.registerEventSource(eventSourceDefinition);
 };
 
+const currencies = new Map([
+    ["USD", "$"],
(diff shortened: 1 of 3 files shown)


ebiggz (approved):
Looks good. Huge bummer that the test data is different from the live data... that is so strange to me.

CaveMobster:
@ebiggz I think it's because the test data also includes support for custom overlay widgets that can be created by the community. But yeah, the documentation is quite unclear about all this unfortunately.

Issue #1458, opened 2021-11-20
[Feature Request] Add ability to resize viewer list and activity feed sidebars in the Dashboard page


CrowbarToolsRobot:
ebiggz via Discord:

Add ability to resize viewer list and activity feed sidebars in the Dashboard page

PR #1565, opened 2022-02-02
Solidify the column resizer

CaveMobster:
### Description of the Change
Making this a PR just to be safe. I solidified the resizer to take into account multiple columns and the variable widths of the parent.

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1458

### Testing
Tried to break it, but didn't succeed.


ebiggz (approved):
Epic

Issue #1025, opened 2021-01-21
[Feature Request] Command Cooldown Groups


CrowbarToolsRobot:
@ebiggz via Discord:

Command Cooldown Groups

PR #1574, opened 2022-02-07
Add an option to add and clear command cooldowns based on their sort tag

CaveMobster:
### Description of the Change
This adds the option to add or clear cooldowns for all commands that have the selected sort tag.

I think this could be pretty useful for toggling triggers too actually. It's like the Event Sets, but with the added option that a trigger can be part of multiple sets. Plus it would keep the code rather simple.

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1025

### Testing
Tested for backwards compatibility, tested the functionality of both single and sort tag cooldowns, and combining cooldown effects.

### Screenshots
image


ebiggz:
@CaveMobster So initially i wanted Sort Tags to be just that - sort tags for the UI. I didn't want to attach other app behaviors to something intended for sorting a table of data. The original thought was if we wanted to have more app behaviors on a "group" of items, we'd integrate the "Sets" (le Event Sets) feature into the firebot-item-table so anything could utilize it. 

At this point though, I don't want to keep forcing that paradigm if you all feel it's unnecessary or too restrictive. So I am fine if we go this route and I am fine with this PR :) I just ask that we at least rename them to "Tags" as they are no longer just for sorting.

CaveMobster:
@ebiggz Ah yep, I 100% understand that approach. It's nice to have things separated and to typically assign one task to one element. However, I do feel tags allow us to integrate flexible behavior across the app based on the data, instead of on the functionality. I think the rename in that sense is good, but maybe naming them "groups" could work even better? A group indicates that its members can be manipulated, whereas a tag still only serves the purpose of categorizing and identifying. An objection I can see there though, is that it's too big a name change and can cause confusion.
What do you think?

CaveMobster:
Actually, scratch that. In their current state they aren't groups, so Tags is a good word. I'll get the change done.

CaveMobster:
@ebiggz The renaming is done.

ebiggz (approved):

Issue #1578, opened 2022-02-10
[Bug] StreamElements donation name is undefined


CrowbarToolsRobot:
**Original Discord message by zunderscore:**
Looks like there's an issue with StreamElements donations where the name is coming in as undefined. Maybe the payload changed and the name field changed?

> Issue created by CaveMobster via Discord

PR #1579, opened 2022-02-10
Fix StreamElements donation "from" name

zunderscore:
This should fix the issue in #1578

### Description of the Change
Change StreamElements donation event trigger to set **from** as the **username** field from the event payload, instead of the non-existent **displayName** field

### Applicable Issues
#1578

### Testing
Verified payload data from debug log

### Screenshots
N/A


CaveMobster (approved):
Awesome, thank you for fixing this!

PR #1594, opened 2022-02-18
Create a view hours importer for Streamlabs Chatbot

CaveMobster:
### Description of the Change
This adds an import modal to the Viewers tab, with which viewers + their view hours can be imported from Streamlabs Chatbot. 

### Testing

Tested the settings and whether the importable viewers get importer correctly.

### Screenshots

image


backend/import/third-party/streamlabs-chatbot.js
```diff
@@ -0,0 +1,122 @@
...
+    let twitchViewers = [];
+
+    const nameGroups = [];
+    if (viewers.length > 100) {
+        while (viewers.length > 100) {
+            nameGroups.push(viewers.splice(0, 100));
+        }
+    } else {
+        nameGroups.push(viewers);
+    }
```
pirak, line 19:
```suggestion
    while (viewers.length > 0) {
        nameGroups.push(viewers.splice(0, 100));
    }
```
I think this would not import some viewers. E.g., for 105 viewers the while loop would be executed only once with `viewers` still containing the last five elements, then jumping to line 21.

`splice` should only take as many elements as possible from the array anyway if I understand  [the documentation for the second parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) correctly, so the if/else shouldn’t be necessary.

ebiggz (approved):

Issue #1613, opened 2022-03-07
[Feature Request] Send custom args to preset effect list API endpoint via GET request query string


zunderscore:
**Describe the solution you'd like**
When triggering a preset effect list via a GET request to its API endpoint (e.g. http://localhost:7472/api/v1/effects/preset/aca8ca70-9e2e-11ec-8d9d-afcfc391f5a8), allow args to be sent via query string parameters (e.g. http://localhost:7472/api/v1/effects/preset/aca8ca70-9e2e-11ec-8d9d-afcfc391f5a8?someValue=value)

**Additional context**
We currently add args and username via POST request body but we don't support via query string on GET requests. This would enable additional functionality via integrations that only support GET and not POST (like Stream Deck's open website as GET request in background feature).

PR #1614, opened 2022-03-07
Add query param handling for preset effect list GET requests

zunderscore:
### Description of the Change
Username and args sent via query string are now passed to a preset effect list when API is hit via GET request.

### Applicable Issues
#1613

### Testing
Created a simple preset effect list that performs TTS using both an arg and the username. Verified that GET requests now use the query string and POST requests are still able to use the request body.

### Screenshots
N/A

### DIFF
--- a/server/api/v1/controllers/effectsApiController.js
+++ b/server/api/v1/controllers/effectsApiController.js
@@ -75,11 +75,18 @@ exports.runPresetList = async function(req, res) {
 
     const body = req.body || {};
     const query = req.query || {};
-    let { args, username } = body;
-
-    if (req.method === "GET") {
+    let args, username;
+    
+    // GET
+    if (req.method === "GET")
+    {
         username = query.username;
         args = query;
+
+    // POST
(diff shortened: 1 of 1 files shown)


server/api/v1/controllers/effectsApiController.js
```diff
@@ -74,7 +74,13 @@ exports.runPresetList = async function(req, res) {
...
 
     const body = req.body || {};
-    const { args, username } = body;
+    const query = req.query || {};
+    let { args, username } = body;
+
+    if (req.method === "GET") {
+        username = query.username;
+        args = query;
+    }
```
SReject, line 83:
```js
let args, username;
if (req.method === "GET") {
  username = query.username;
  args = query;
} else {
  username = body.username;
  args = body.args
}

SReject (changes requested):

SReject (approved):

ebiggz (approved):

Issue #1621, opened 2022-03-11
[Feature Request] Get viewer metadata via API


zunderscore:
**Describe the solution you'd like**
When querying the Firebot API at `/api/v1/viewers/:userId`, I'd like to get all available user metadata.

**Additional context**
This is to expand on feature request #1558. This may include any user metadata such as total view time, total chat messages, currency totals, etc.

PR #1633, opened 2022-03-23
Get user metadata via API

zunderscore:
### Description of the Change
Added ability to get user metadata via API. Usage is either `api/v1/viewers/[user_id]` or `api/v1/viewers/[username]?username=true`, including all data in the FirebotUser object that the user manager returns, currencies, and custom viewer roles.

### Applicable Issues
#1621

### Testing
Accessed several users' metadata via API endpoint

### Screenshots
N/A


server/api/v1/controllers/viewersApiController.js
```diff
@@ -1,12 +1,41 @@
...
+    if (userId == null) {
+        return res.status(400).send({
+            status: "error",
+            message: `No viewerIdOrName provided`
+        });
+    }
+
+    const metadata = username === "true" ?
+        (await userDb.getUserByUsername(userId)) :
+        (await userDb.getUserById(userId));
```
SReject, line 24: If blocks are preferred when using await

```js
let metadata;
if (username === "true") {
  metadata = await userDb.getUserByUsername(userId);
} else {
  metadata = await userDb.getUserByUsername(userId);
}
```

SReject (changes requested):

server/api/v1/controllers/viewersApiController.js
zunderscore: Done

SReject (approved):

Issue #1618, opened 2022-03-09
[Feature Request] Add $keys[] variable to extract object keys


CrowbarToolsRobot:
SReject via Discord:

Add a variable that returns an array of an object's own keys. This is mostly related to custom variables

CKY-:
closing as released

PR #1637, opened 2022-03-24
Add $customVarialbeKeys[] variable to extract object keys

Samuel38:
### Description of the Change
This adds a $customVarialbeKeys[variableName] variable that extracts the array of keys of a custom variable. 

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1618

### Testing
Created a custom command to create custom variables and output the result of the custom variables keys. 
Also successfully tried to loop over the resulting array of keys. 

### Screenshots
test_keys

### DIFF
--- a/backend/variables/builtin/custom-variable-keys.js
+++ b/backend/variables/builtin/custom-variable-keys.js
@@ -1,0 +1,40 @@
+"use strict";
+
+const customVariableManager = require("../../common/custom-variable-manager");
+
+const { OutputDataType, VariableCategory } = require("../../../shared/variable-constants");
+
+function isObject(data) {
+	return typeof data === 'object' && !(typeof data === 'string' || data instanceof String);
+}
+
+const model = {
+	definition: {
+		handle: "customVariableKeys",
+		usage: "customVariableKeys[name]",
+		examples: [
+			{
+				usage: "customVariableKeys[name, 1]",
+				description: "Get the array of keys for an object which is an array item by providing an array index as a second argument."
+			},
+			{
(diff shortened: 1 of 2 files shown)


backend/variables/builtin/custom-variable-keys.js
```diff
@@ -0,0 +1,40 @@
...
+				usage: "customVariableKeys[name, property]",
+				description: "Get the array of keys for an object property by providing a property path (using dot notation) as a second argument."
+			}
+		],
+		description: "Get the array of keys for an object saved in the custom variable.",
+		categories: [VariableCategory.ADVANCED],
+		possibleDataOutput: [OutputDataType.TEXT]
+	},
+	evaluator: (_, name, propertyPath) => {
+		let data = customVariableManager.getCustomVariable(name, propertyPath);
```
ebiggz, line 30: Super small thing but we can use `const` instead of `let` since it doesn't get reassigned and we don't expect it to.

ebiggz (changes requested):
Great work! Thanks for taking the time to implementing this. One super small piece of feedback otherwise looks good to go 🎉

Samuel38:
That's a good point. I made the change.

SReject (changes requested):
We use 4 spaces for indents. and line-feed(`\x0A`) for end-of-line

SReject:
A big issue that we should have spotted sooner:

Master branch is used for current latest release. All dev work happens against the v5 branch. You will need to fork from the v5 branch and move your work over to it.

PR #1642, opened 2022-03-26
Added displayName to fix #1607

### LINKED ISSUE #1607: [Bug] Chat tag ping no triggering
Perry via Discord:

If someone tags the streamer `@UserName` it doesn't trigger, if you however do `@username` it triggers.

brumoen:
### Description of the Change
Added check for `streamer.displayName` to username tag check for Chat Notification so that it triggers on both names and not only the username.

### Applicable Issues
#1607 

### Testing
Manually tested that the bot notifies on both variants of the username

### DIFF
--- a/backend/chat/chat-helpers.js
+++ b/backend/chat/chat-helpers.js
@@ -204,7 +204,7 @@ function parseMessageParts(firebotChatMessage, parts) {
                 if (!firebotChatMessage.whisper &&
                 !firebotChatMessage.tagged &&
                 streamer.loggedIn &&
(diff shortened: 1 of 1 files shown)


zunderscore:
Would it make sense to make those checks case insensitive, just to be able to catch any variant of the username/displayName?

For example, my username and display name are both formatted as **zunderscore**, but if someone goes "ZUNDERSCORE" in my chat, it wouldn't be highlighted, but I think I would want it to.

SReject (approved):

Issue #1638, opened 2022-03-24
[Bug] ArrayFind result on an array of strings should return index


Samuel38:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We create custom variable 
`test` with content `["test1", "test2", "test3"]`
Using `$arrayFind[test, "test2"]` will return `test2`
Using `$arrayFind[test, "test4"]` will return `null`

### Expected Behavior

Using `$arrayFind[test, "test2"]` should return `2`
Using `$arrayFind[test, "test4"]` should return `null`

In the current state, running `$arrayFind` on an array of strings doesn't return relevant information. 
While we would be looking at the location of the string in the array, it only returns the string we have input if it existed. So aside from telling us that the array actually contains that string, it doesn't say anything. 

### Version

5.50.4

### What operating system are you using Firebot on?

Windows 10

### Contact Details (Optional)

Alastor#8593 in discord

ebiggz:
Hi @Samuel38! Thanks for the report. 

The current behavior is actually indeed the intended behavior. The biggest use case is most evident with the second example provided in the app when you are looking to find objects in an array:
```
$arrayFind["[{\\"username\\": \\"ebiggz\\"},{\\"username\\": \\"MageEnclave\\"}]", ebiggz, username]
```
Imagine you have an array of user data you are tracking and you want to get the object for a specific person. You can use their username to find the full object. This behavior is intended to mirror the `array.find` functionality in javascript.

The functionality you expect is valid and would be useful too though. So perhaps we create a secondary variable for that functionality ie:
```
$arrayFindIndex[...]
```

Samuel38:
Hi, 
Thanks for your answer. 

I do see the use case in extracting objects from an array, and I'm not suggesting to change that one. 
My thought was that what is returned in the case where no property path is given makes it useless. Adding another function would be fine too, but would have the drawback of multiplying functions and maybe overwhelming a little bit the users that are less experienced in programming. 
On the other hand, my suggestion makes the behaviour of this function less consistent. 
I guess a middle ground could be to add an optionnal argument to explicitly ask for the index `$arrayFind[name, match, path=null, giveIndex=false]`.

PR #1644, opened 2022-03-26
#1638 make $arrayFind return index

Samuel38:
### Description of the Change
Added a new boolean optional parameter to return the index instead of the object. 
New usage : $arrayFind[jsonArray, matcher, propertyPath, returnIndex]
Also proposed an update to the descriptions to clarify the behaviour. 

This change goes along the lines of my last proposition in the bug report thread. 
Rationale : 
- Changing the behavior to return index just for arrays of non objects would be inconsistant and possibly break existing setups
- Creating an additional function would add clutter and more easily overwhelm the user base that's not used to programming
- Adding an optional parameter keeps current behavior unchanged while expanding the capabilities. 

### Applicable Issues
#1638 

### Testing
Created a chat command with the following : 
$arrayFind["[\"a\",\"b\",\"c\"]", b] | 
$arrayFind["[{\"username\": \"ebiggz\"},{\"username\": \"MageEnclave\"}]", ebiggz, username] | 
$arrayFind["[\"a\",\"b\",\"c\"]", b, null, true] | 
$arrayFind["[{\"username\": \"ebiggz\"},{\"username\": \"MageEnclave\"}]", ebiggz, username, true] || 
$arrayFind["[\"a\",\"b\",\"c\"]", d]  | 
$arrayFind["[{\"username\": \"ebiggz\"},{\"username\": \"MageEnclave\"}]", toto, username] | 
$arrayFind["[\"a\",\"b\",\"c\"]", d, null, true] | 
$arrayFind["[{\"username\": \"ebiggz\"},{\"username\": \"MageEnclave\"}]", toto, username, true]

Also tested a number of them by setting a customVariable and requesting it back

### Screenshots
test_arrayFind

### DIFF
--- a/backend/variables/builtin/array-find.js
+++ b/backend/variables/builtin/array-find.js
@@ -24,16 +24,20 @@ function getPropertyAtPath(obj, propertyPath) {
 const model = {
     definition: {
         handle: "arrayFind",
-        usage: "arrayFind[jsonArray, matcher, propertyPath]",
-        description: "Finds a matching element in the array or null",
+        usage: "arrayFind[jsonArray, matcher, propertyPath, returnIndex]",
+        description: "Finds a matching element in the array or null. Returns either the object index or the object",
         examples: [
             {
-                usage: 'arrayFind["[1,2,3]", 1]',
-                description: "Finds 1 in the array"
+                usage: 'arrayFind["["a","b","c"]", "b"]',
+                description: 'Returns "b"'
             },
             {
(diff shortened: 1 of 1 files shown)


SReject (changes requested):
This breaks backwards compatibility with those that use `$arrayFind[]`'s current behavior. Instead, create a new variable `$arrayFindIndex[]`

SReject:
A big issue that we should have spotted sooner:

Master branch is used for current latest release. All dev work happens against the v5 branch. You will need to fork from the v5 branch and move your work over to it.

PR #1647, opened 2022-03-27
Fix github link

quack79:
There was an additional /Firebot in the URL, resulting in a 404

### Description of the Change
Removed part of the URL from the Github link in the About window, to point at the correct location of the source.


SReject:
Changes should be made against the `v5` branch, not `master`

PR #1648, opened 2022-03-27
#1618 add $customVarialbeKeys variable

Samuel38:
### Description of the Change
This adds a $customVarialbeKeys[variableName] variable that extracts the array of keys of a custom variable. 
This is a resubmit of PR #1637 with fixed issues and onto the correct branch. 
Per the comment about formatting, I couldn't find any line-feed issue in my commits. So hopefully it's fine. I fixed the tabs. 

### Applicable Issues
#1618

### Testing
Created a custom command to create custom variables and output the result of the custom variables keys. 
Also successfully tried to loop over the resulting array of keys. 

### Screenshots
test_keys

### DIFF
--- a/backend/variables/builtin/custom-variable-keys.js
+++ b/backend/variables/builtin/custom-variable-keys.js
@@ -5,35 +5,35 @@ const customVariableManager = require("../../common/custom-variable-manager");
 const { OutputDataType, VariableCategory } = require("../../../shared/variable-constants");
 
 function isObject(data) {
-	return typeof data === 'object' && !(typeof data === 'string' || data instanceof String);
+    return typeof data === 'object' && !(typeof data === 'string' || data instanceof String);
 }
 
 const model = {
-	definition: {
-		handle: "customVariableKeys",
-		usage: "customVariableKeys[name]",
-		examples: [
-			{
-				usage: "customVariableKeys[name, 1]",
-				description: "Get the array of keys for an object which is an array item by providing an array index as a second argument."
-			},
-			{
-				usage: "customVariableKeys[name, property]",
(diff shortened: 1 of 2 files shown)


backend/variables/builtin/custom-variable-keys.js
```diff
@@ -0,0 +1,40 @@
+"use strict";
+
+const customVariableManager = require("../../common/custom-variable-manager");
+
+const { OutputDataType, VariableCategory } = require("../../../shared/variable-constants");
+
+function isObject(data) {
+    return typeof data === 'object' && !(typeof data === 'string' || data instanceof String);
```
SReject, line 8: Can be simplified to:

```js
return typeof data === 'object' && !(data instanceof String)
```

If `data` returns truthy against the `typeof data === 'object'` check, it won't ever pass the `=== 'string'` check

backend/variables/builtin/custom-variable-keys.js
```diff
@@ -0,0 +1,40 @@
...
+                usage: "customVariableKeys[name, property]",
+                description: "Get the array of keys for an object property by providing a property path (using dot notation) as a second argument."
+            }
+        ],
+        description: "Get the array of keys for an object saved in the custom variable.",
+        categories: [VariableCategory.ADVANCED],
+        possibleDataOutput: [OutputDataType.TEXT]
+    },
+    evaluator: (_, name, propertyPath) => {
+        const data = customVariableManager.getCustomVariable(name, propertyPath);
```
SReject, line 30: Add null check after to exit early
```js
if (data == null) {
  return "[]";
}
```

SReject (changes requested):
Changes requested

Samuel38:
Changes applied

SReject (approved):

PR #1649, opened 2022-03-27
#1638 add $arrayFindIndex variable

Samuel38:
### Description of the Change
This adds a $arrayFindIndex[jsonArray, matcher, propertyPath] variable that extracts the index of the array `jsonArray` where the item matches `matcher`. If `propertyPath` is filled, it tries to match the object item by following the property path. 
If it can't find a match, it returns null. 

### Applicable Issues
#1638 

### Testing
Created a chat command with the following : 
$arrayFindIndex["[\"a\",\"b\",\"c\"]", b] | 
$arrayFindIndex["[{\"username\": \"alastor\"},{\"username\": \"ebiggz\"}]", alastor, username] || 
$arrayFindIndex["[\"a\",\"b\",\"c\"]", d]  | 
$arrayFindIndex["[{\"username\": \"alastor\"},{\"username\": \"ebiggz\"}]", toto, username]

Checked the return value is "1 | 0 || null | null"

### Screenshots
test_arrayFindIndex

### DIFF
--- a/backend/variables/builtin/array-find-index.js
+++ b/backend/variables/builtin/array-find-index.js
@@ -48,7 +48,7 @@ const model = {
             try {
                 matcher = JSON.parse(matcher);
             } catch (err) {
-                //fail silently
+                return null;
             }
 
             if (propertyPath === 'null' || propertyPath === "") {
@@ -74,7 +74,7 @@ const model = {
                     return JSON.stringify(found != -1 ? found : null);
                 }
             } catch (error) {
-                // fail silently
+                return null;
             }
         }
         return null;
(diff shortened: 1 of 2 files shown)


backend/variables/builtin/array-find-index.js
```diff
@@ -0,0 +1,84 @@
...
+    evaluator: (_, jsonArray, matcher, propertyPath = null) => {
+        if (jsonArray != null) {
+            if (matcher === undefined || matcher === "") {
+                return null;
+            }
+
+            try {
+                matcher = JSON.parse(matcher);
+            } catch (err) {
+                //fail silently
```
SReject, line 51: return `null` instead of silently failing

Saves a bit of cycles needless-processing

SReject (changes requested):

Samuel38:
Change applied in both locations where applicable

SReject (approved):

Issue #1597, opened 2022-02-21
[Feature Request] Programmatic access to adding quotes 


TioMecanico:
**Describe the solution you'd like**
A way to add quotes to the quote system without going through the standard !quote command

**Additional context**
Two examples of ways it could be used:
A channel point redemption that adds a quote to the list
A custom command that adds a quote to the list with an additional effects as well

SReject:
For clarity, do you mean something of a "Add Quote" Effect?

TioMecanico:
Yeah, basically. Something that takes input (either in one or two arguments) and acts as if someone had done `!quote add` with them, but without using the `!quote add` Command.

zunderscore:
@TioMecanico, with the stuff added between the Add Quote effect and the new API endpoints, does that satisfy what you were looking for here? If not, what else would help?

SReject:
Resolved with v5.52.0

PR #1660, opened 2022-04-01
Quote API endpoints

zunderscore:
### Description of the Change
This adds quote access via the internal API. The following endpoints have been added:

`/api/v1/quotes`
**GET**: Gets all quotes
**POST**: Adds a new quote

`/api/v1/quotes/:quoteId`
**GET**: Gets quote at specified ID
**PUT**: Adds or updates quote with specified ID
**PATCH**: Updates quote with field values provided in body
**DELETE**: Removes quote with specified ID

Any IDs returned from API are formatted as `id` instead of internal `_id` for clarity.

### Applicable Issues
#1597

### Testing
- Performed GET on `/api/v1/quotes` and received all quotes
- Performed POST on `/api/v1/quotes` and got back inserted quote with new ID
- Performed GET on `/api/v1/quotes/:quoteId` with both valid and invalid quote IDs
- Performed PUT on `/api/v1/quotes/:quoteId` with both new and existing quote IDs. New quote ID which was higher than previous autoincrement successfully set new autoincrement value.
- Performed PATCH on `/api/v1/quotes/:quoteId` with multiple variations of patch data. All properly specified fields were updated in quotes, invalid values were ignored.
- Performed DELETE on `/api/v1/quotes/:quoteId` and quote was deleted

### Screenshots
N/A


backend/quotes/quotes-manager.js
```diff
@@ -57,22 +70,33 @@ function setQuoteIdIncrementer(number) {
...
 function addQuote(quote) {
     return new Promise(async (resolve, reject) => {
-
-        let newQuoteId = await getNextQuoteId();
-        if (newQuoteId == null) {
-            logger.error("Unable to add quote as we could not generate a new ID");
-            return reject();
+        // If no/invalid ID is specified, get the next one
+        if (!quote._id || isNaN(quote._id)) {
+            let newQuoteId = await getNextQuoteId();
```
ebiggz, line 75: Looks like this `let` never gets reassigned and so could be a `const` instead

server/api/v1/controllers/quotesApiController.js
```diff
@@ -0,0 +1,279 @@
...
+        return res.status(500).send({
+           status: "error",
+           message: "Error creating quote: " + e
+        });
+    }
+};
+
+exports.putQuote = async function(req, res) {
+    let { quoteId } = req.params;
+    let quotePut = req.body;
```
ebiggz, line 80: Another `let` that I believe can be a `const` as well

server/api/v1/controllers/quotesApiController.js
```diff
@@ -0,0 +1,279 @@
...
+        creator: quote.creator,
+        game: quote.game,
+        createdAt: quote.createdAt
+    };
+}
+
+// Validates that all required quote fields are present and valid
+function validateQuote(quote) {
+    quote = quote ?? {};
+    let validationErrors = [];
```
ebiggz, line 268: `const` here too.

server/api/v1/controllers/quotesApiController.js
```diff
@@ -0,0 +1,279 @@
...
+            message: "Invalid quoteId provided"
+        });
+    }
+
+    const quote = await quotesManager.getQuote(quoteId);
+
+    if (quote == null) {
+        return res.status(404).send({
+            status: "error",
+            message: "Quote " + quoteId + " not found"
```
ebiggz, line 39: Although a lot of our legacy code is bad about this, for new stuff we generally want to always use string template literals instead of string concatenation. IE:
```js
{
  //...
  message: `Quote ${quoteId} not found`
}
```

Looks like there's a few places here that an be updated to template literals.

server/api/v1/controllers/quotesApiController.js
```diff
@@ -0,0 +1,279 @@
...
+        });
+    }
+
+    quoteId = parseInt(quoteId);
+    if (isNaN(quoteId)) {
+        return res.status(400).send({
+            status: "error",
+            message: "Invalid quoteId provided"
+        });
+    }
```
ebiggz, line 32: The logic for these two checks are reused for a number of the endpoints here.  Good candidate to bump out into a function to keep us DRY.

Perhaps something like
```js
function validateQuoteId(quoteId, res) {
   if (quoteId == null) {
        res.status(400).send({
            status: "error",
            message: "No quoteId provided"
        });
        return false;
    }

    quoteId = parseInt(quoteId);
    if (isNaN(quoteId)) {
        res.status(400).send({
            status: "error",
            message: "Invalid quoteId provided"
        });
        return false;
    }
    
    return quoteId;
}

// elsewhere...
quoteId = validateQuoteId(quoteId, res);
if(!quoteId) {
  return;
}
```

server/api/v1/controllers/quotesApiController.js
```diff
@@ -0,0 +1,279 @@
...
+        quote.creator = quotePatch.creator.replace(/@/g, "");
+    }
+
+    if (quotePatch.game != null && quotePatch.game !== "") {
+        quote.game = quotePatch.game;
+    }
+
+    if (quotePatch.createdAt && quotePatch.createdAt != null && quotePatch.createdAt !== "") {
+        quote.createdAt = moment(quotePatch.createdAt).toISOString();
+    }
```
ebiggz, line 207: We can use optional chaining to simply these null and empty checks. This feature wasn't always available for Firebot until more recently so a lot of our code doesn't use it yet but definitely worth doing so going forward.

EX:
```js
if (quotePatch.text?.length > 0) {
        quote.text = quotePatch.text;
}
```

server/api/v1/controllers/quotesApiController.js
```diff
@@ -0,0 +1,279 @@
...
+    quote = quote ?? {};
+    let validationErrors = [];
+    
+    if (!quote.text || quote.text == null || quote.text === "") {
+        validationErrors.push("Missing quote text");
+    }
+    
+    if (!quote.creator || quote.creator == null || quote.creator === "") {
+        validationErrors.push("Missing quote creator");
+    }
```
ebiggz, line 276: Another good opportunity for optional chaining here. `if(quote.text?.length < 1) { ... }`, etc

ebiggz (changes requested):
Well done! By and large looks good, just some various stylistic suggestions.

backend/quotes/quotes-manager.js
zunderscore: Fixed

server/api/v1/controllers/quotesApiController.js
zunderscore: Fixed

server/api/v1/controllers/quotesApiController.js
zunderscore: Fixed

server/api/v1/controllers/quotesApiController.js
zunderscore: Fixed multiple

server/api/v1/controllers/quotesApiController.js
zunderscore: Fixed

server/api/v1/controllers/quotesApiController.js
zunderscore: Fixed. Negative checking is a little weirder with optional chaining, so it was either something like `if (!quote.text || quote.text?.length < 1)` or `if (!(quote.text?.length > 0)` so I opted for the latter to be concise.

ebiggz (approved):
Looks epic!

Issue #1662, opened 2022-04-02
[Bug] Community gifted subs trigger individual sub notifications intermittently 


mgway:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a viewer gifts X random subscriptions to the community, sometimes the individual gift sub event fires X times and sometimes the community gift sub event is fired. The occurrence of this issue is independent of the number of subs gifted but appears strongly correlated to a protracted delay between the "UserName is gifting X subscriptions to the community!" header message and the list of recipients appearing in chat. 

Related: 
https://github.com/crowbartools/Firebot/issues/1521
https://discord.com/channels/372817064034959370/372818514312167424/927512326117138463

### Expected Behavior

Community gifted subs always trigger the community gifted sub alert, not the individual sub alert. 

### Steps To Reproduce

- Gift community subs and hope the stars align
- This appears easier to reproduce when the broadcaster is offline so Twitch has to "work harder" to find gift recipients across followers

### Version

5.51.0

### What operating system are you using Firebot on?

Windows 10

### Relevant log output

```shell
16:24:47 GMT-0400 (Eastern Daylight Time) - debug: [v5.51.0] Received 5 community gift subs from An Anonymous Gifter at 16:24:47:66
16:24:49 GMT-0400 (Eastern Daylight Time) - debug: [v5.51.0] Running next effects for queue f7c9af30-870c-11ec-9163-f5490c75f692. Mode=custom, Interval?=10, Remaining queue length=0
16:24:49 GMT-0400 (Eastern Daylight Time) - debug: [v5.51.0] Looking and handling replace variables...
16:24:49 GMT-0400 (Eastern Daylight Time) - debug: [v5.51.0] Running firebot:playsound(41d6b240-a1d8-11ec-963f-8ba0dca942b1) effect...
16:24:49 GMT-0400 (Eastern Daylight Time) - debug: [v5.51.0] Deleting token: j2a7cq89fwjmnsveb5mqk
16:24:49 GMT-0400 (Eastern Daylight Time) - debug: [v5.51.0] Looking and handling replace variables...
16:24:49 GMT-0400 (Eastern Daylight Time) - debug: [v5.51.0] Running firebot:playvideo(41d6b241-a1d8-11ec-963f-8ba0dca942b1) effect...
16:24:49 GMT-0400 (Eastern Daylight Time) - debug: [v5.51.0] Looking and handling replace variables...
[...]

PR #1663, opened 2022-04-02
#1662 Use Chat interface for gift sub events

mgway:
### Description of the Change
- Use chat interface for handling sub gift events
- Increase community gift cache TTL to 10 seconds.

These two changes reduce the strict ordering requirement that arose from using both PubSub and Chat interfaces for handling community gifted subs. In the current system, the ChatCommunitySubInfo chat event MUST arrive between 0 and 3 seconds before any ChatSubGiftInfo PubSub events. Since these are two different transport mechanisms with differing latency, there is no explicit guarantee that the events will arrive in the correct order. Further, with large numbers of gifted subscriptions, Twitch has to "think about it" for a while, with the "User X gifted Y subs to the community" message appearing >>3 seconds ahead of the list of recipient names.

In summary, this proposed change switches gift sub event handling to the chat interface instead of PubSub to handle the event ordering issue and increases the cache TTL to 10 seconds to handle Twitch-side delays in issuing the list of recipient names. 

### Applicable Issues
#1662

### Testing
Tested gifting two subs to the community. Before this fix, the delay of 5 seconds between `Received 2 community gift subs from Celina` and `Attempting to process community gift sub from Celina` would have resulted in individual alerts being fired as the 3 second TTL `communitySubCache` entry would have expired before the first gift recipient message appeared in chat/was received by Firebot. 

```
19:19:00 GMT-0700 (Pacific Daylight Time) - debug: [v5.51.43] ViewerDB: Setting last seen date for 25 users
19:19:06 GMT-0700 (Pacific Daylight Time) - debug: [v5.51.43] Received 2 community gift subs from Celina at 19:19:06:88
19:19:11 GMT-0700 (Pacific Daylight Time) - debug: [v5.51.43] Attempting to process community gift sub from Celina at 19:19:11:31
19:19:11 GMT-0700 (Pacific Daylight Time) - debug: [v5.51.43] Attempting to process community gift sub from Celina at 19:19:11:32
19:19:11 GMT-0700 (Pacific Daylight Time) - debug: [v5.51.43] Community gift sub event triggered, deleting cache
19:19:11 GMT-0700 (Pacific Daylight Time) - debug: [v5.51.43] Sending effects for list a4d92d35-a4cd-11ec-8c2f-e7dede52b7c9 to queue 906155b0-87d3-11ec-9dd4-d10d385802f1...
```


ebiggz (approved):
Thanks for doing this. We originally used the chat interface for sub but tried to migrate as much events as we could to pubsub. This one was always a problem child for all the reasons you detailed however so this is definitely the right move.

PR #1675, opened 2022-04-06
Get all preset effect lists via API

zunderscore:
### Description of the Change
Adds an API endpoint at `api/v1/effects/preset` to GET all preset effect lists, including ID, name, and a list of arguments they take

### Applicable Issues
N/A

### Testing
GET-ted the list. Hooray!

### Screenshots
N/A

### DIFF
--- a/server/api/v1/controllers/effectsApiController.js
+++ b/server/api/v1/controllers/effectsApiController.js
@@ -55,6 +55,27 @@ exports.runEffects = async function(req, res) {
     }
 };
 
+exports.getPresetLists = async function(req, res) {
+    const presetLists = presetEffectListManager.getAllItems();
+
+    if (presetLists == null) {
+        return res.status(500).send({
+            status: "error",
+            message: "Unknown error getting preset effect lists"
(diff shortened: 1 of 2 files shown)


server/api/v1/controllers/effectsApiController.js
```diff
@@ -55,6 +55,27 @@ exports.runEffects = async function(req, res) {
...
+        });
+    }
+
+    const formattedPresetLists = presetLists.map(l => {
+        return {
+            id: l.id,
+            name: l.name,
+            args: l.args.map(a => a.name)
+        };
+    });
```
SReject, line 74: Use more descriptive names. `listItem` and `itemArg` would suffice

SReject (changes requested):

server/api/v1/controllers/effectsApiController.js
zunderscore: Slightly different names, but done.

SReject (approved):

PR #1676, opened 2022-04-06
Basic Command API endpoints

zunderscore:
### Description of the Change
Adds the following API endpoints for commands:

`/api/v1/commands/system`
- GET list of all system command IDs, names, and base triggers

`/api/v1/commands/system/:sysCommandId`
- GET full definition for specified system command

`/api/v1/commands/system/:sysCommandId/run`
- GET or POST to run system command. Accepts parameters of `username`, `args`, and (POST only) trigger `metadata` object

`/api/v1/commands/custom`
`/api/v1/commands/custom/:customCommandId`
`/api/v1/commands/custom/:customCommandId/run`
Same as above, but for user-defined custom commands. Note that custom command list returns description instead of name.

**NOTE:** This currently executes a command using the same method call as the **Run Command** effect, so subcommand triggers are not supported, just as in the effect.

### Applicable Issues
N/A

### Testing
Tested against all endpoints, verified data and also command execution where applicable.

### Screenshots
N/A

### DIFF
--- a/backend/effects/builtin/log-message.js
+++ b/backend/effects/builtin/log-message.js
@@ -38,7 +38,7 @@ const addFirebotLogMessage = {
     },
     optionsValidator: effect => {
         let errors = [];
-        if (!(effect.logMessage?.length > 1)) {
+        if (!(effect.logMessage?.length > 0)) {
             errors.push("Please input a log message.");
         }
         if (effect.logLevel == null) {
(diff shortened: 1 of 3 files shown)


SReject (changes requested):
Use more descriptive names in functional loops.

SReject (approved):

Issue #922, opened 2020-07-28
[Feature Request] Add the ability to filter specific users from the viewer list.


Sam-Beynon:
@TwinklingPlays I would very much be able to hide users from appearing.. no bans, no nothing.. just hiding them, or categorizing them to "bots" and they show at the bottom of the viewerlist. i know they arent harmful, im just tired of seeing their names in my chat when i know they do 0 for me.

MageEnclave:
To be honest, I think the viewer list is extremely low priority. The viewer list is something that, while sometimes useful, doesn't have much of an impact on the most important pieces of Firebot. I just want to be transparent in saying that it may be a bit before we get to adding it back, much less being able to filter it like this.

Firebottle:
PR has been opened for this issue. It will allow you to hide users via their user modal.

TwinklingPlays:
> 

awesome - twitchinsights keeps their bot list updated. That could be used to check usernamed on there and then categorize them in firebot based on that.

TwinklingPlays:
No hiding pls - or settings at the very least.

PR #1737, opened 2022-05-12
feat: ability to filter users from the viewer list

Firebottle:
### Description of the Change
This implements #922. It gives users the ability to hide usernames from the viewer list that sits next to chat. This can be toggled on per user via the user modal.

When the setting is updated, the viewer will remain on the viewer list until the next time that viewer list refresh event occurs, at which time they will be filtered out.

### Applicable Issues
#922 

### Testing
- Connect firebot and let users show in the userlist.
- Click a username (preferably a persistent bot account) to open their modal.
- Click the new hide from viewer list option.
- Close the modal.
- Wait until the next chat refresh happens, and observe that the username disappears.
- Go into the viewer database and find the name again.
- Click it to open the viewer modal.
- Deselect the new option.
- Go back to chat and wait for the refresh, then observe the username comes back.

### Screenshots
image

### DIFF
--- a/backend/chat/chat-listeners/active-user-handler.js
+++ b/backend/chat/chat-listeners/active-user-handler.js
@@ -96,7 +96,6 @@ async function updateUserOnlineStatus(userDetails, updateDb = false) {
         onlineUsers.set(userDetails.id, true, ONLINE_TIMEOUT);
 
         const roles = await chatRolesManager.getUsersChatRoles(userDetails.id);
-        const details = await userDatabase.getUserById(userDetails.id);
 
         frontendCommunicator.send("twitch:chat:user-joined", {
             id: userDetails.id,
@@ -104,7 +103,7 @@ async function updateUserOnlineStatus(userDetails, updateDb = false) {
             roles: roles,
             profilePicUrl: userDetails.profilePicUrl,
             active: exports.userIsActive(userDetails.id),
-            disableViewerList: details.disableViewerList ? details.disableViewerList : false
+            disableViewerList: userDetails.disableViewerList
         });
 
         if (updateDb) {
@@ -134,7 +133,8 @@ exports.addOnlineUser = async (username) => {
                 username: twitchUser.name,
                 displayName: twitchUser.displayName,
                 twitchRoles: [],
-                profilePicUrl: twitchUser.profilePictureUrl
+                profilePicUrl: twitchUser.profilePictureUrl,
+                disableViewerList: false
             };
 
             chatHelpers.setUserProfilePicUrl(twitchUser.id, twitchUser.profilePictureUrl);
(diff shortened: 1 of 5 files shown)


backend/chat/chat-listeners/active-user-handler.js
```diff
@@ -96,13 +96,15 @@ async function updateUserOnlineStatus(userDetails, updateDb = false) {
         onlineUsers.set(userDetails.id, true, ONLINE_TIMEOUT);
 
         const roles = await chatRolesManager.getUsersChatRoles(userDetails.id);
+        const details = await userDatabase.getUserById(userDetails.id);
```
Firebottle, line 99: There is a new possibly DB hit here to get user settings from the user modal and add it to the data sent to the front end.

gui/app/directives/chat/chat-user-category.js
```diff
@@ -15,7 +15,7 @@
                 <div style="font-size: 12px; opacity: 0.6;">{{$ctrl.category}}</div>
                 <div
                     class="chat-user-wrapper"
-                    ng-repeat="user in cms.chatUsers | chatUserRole:$ctrl.roleKey | orderBy:'username':true | orderBy:'active':true as filtered track by user.id"
+                    ng-repeat="user in cms.getFilteredChatUserList() | chatUserRole:$ctrl.roleKey | orderBy:'username':true | orderBy:'active':true as filtered track by user.id"
```
Firebottle, line 18: Note the viewer list will always run through this new filter function instead of pulling directly from our main chatUsers array.

gui/app/services/chat-messages.service.js
```diff
@@ -48,6 +48,16 @@
...
 
+            // Return User List with people in role filtered out.
+            service.getFilteredChatUserList = function() {
+                let userList = service.chatUsers;
+                userList = userList.filter(function(user) {
+                    return user.disableViewerList != null ? !user.disableViewerList : true;
+                });
+
+                return userList;
+            };
```
ebiggz, line 59: We can simplify this a little bit with a few small tweaks:
1) forgo the variable assignment
2) use a fat-arrow function for the filter lamda
3) since undefined/null is considered a falsey value in js, we dont need to expressly check for it here
```ts
service.getFilteredChatUserList = function() {
    return service.chatUsers.filter((user) =>!user.disableViewerList);
};
```

backend/chat/chat-listeners/active-user-handler.js
ebiggz: This is a good callout and actually might be something we can avoid. Both functions that call `updateUserOnlineStatus` (namely `addOnlineUser` and `addActiveUser`) already get user data from the database, so we could instead make `disableViewerList` an argument of `updateUserOnlineStatus` and pass the bool in from those other funcs.

Additionally, a cool little trick to ensure a value is a boolean is to use a double-bang, like so: `!!details.disableViewerList`. Its functionally equivalent to `details.disableViewerList ? details.disableViewerList : false`

ebiggz (changes requested):
Awesome work, this will be a great feature

ebiggz (approved):
Epic!!

Issue #1733, opened 2022-05-11
[Bug] Commands Currency auto deduction is being removed while command is on cooldown 


CKY-:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

create a command  
use the currency  restriction with 
Automatically deduct currency from user if restrictions pass enabled 
set a cooldown user and global 
while the command is on cooldown the command still Automatically deducts currency from user

CC:discord via Shen Nung 
Hello, I'm trying to set up a currency system where viewers can purchase an image and it stays on screen for the duration of the stream.  Each image can only be bought once per stream and I've set the global cooldown to a high number to ensure that.  The restriction used is currency >= cost and I have auto-deduction of currency on.  The only issue I'm having is that if the command is used while on cooldown, it still auto-deducts the currency (the cooldown message appears in chat as normal).  Any suggestions?

### Expected Behavior

do Nothing while on cool down 

### Version

v5.52.0

### What operating system are you using Firebot on?

Windows 10

### Contact Details (Optional)

cky

SReject:
Resolved with release v5.53.0

PR #1741, opened 2022-05-12
fix(chat): :bug: Check for command cooldown before restrictions.

Firebottle:
### Description of the Change
This updates commands so that the cooldown status of the command is checked before the restrictions. Our currency effect has an auto deduct function, and people were passing the restriction and paying the cost, then afterwards they would find out the command was still on cooldown.

The cooldown ultimately overrides all restrictions anyway, so we might as well check it first. I can't think of any situation where this might cause issues with other effects.

### Applicable Issues
#1733 

### Testing
- Create a command with a currency restriction, auto deduct on, and a cooldown.
- Run the command once and make sure currency is deducted.
- Run the command a second time, while cooldown is still active, and make sure currency is not deducted and you get the cool down message.

### Screenshots
N/A


ebiggz (approved):
Thanks for doing this, exactly what needed to happen.

Issue #1766, opened 2022-05-20
[Bug] Disabling an effect in random effect still treats it as an effect that can be chosen


Badhabits:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Will 100% be banned for finding bugs and giving to much work. However your work is so very very appreciated. You have no idea how much this bot has done for me. Sorry again.

If you disable an effect in a random effect it still treats it as an effect that can be triggered, however since it is disabled it will not trigger ,but will not allow something else to be chosen instead.

### Expected Behavior

If you disable an effect within a random effect it should exclude it from being chosen so another effect can be chosen instead. 

Its confusing I'm sorry, steps to reproduce should help.

### Steps To Reproduce

1. Create a command
2. Create an effect: custom var
3. Custom Var = 1
4. create a run random effect
5. Create an effect within the run random effect: custom var
6. Custom Var = 2
7. Duplicate the custom Var effect within the random effect 3 times
8. You should have 4 Custom variable effects within the Run Random effect all with a variable of 2
9. (Toggle Disable) Disable 2 of them
10. Create an effect: Chat effect In the command to output the results of the Var we made.
11.  Run the command 10 times, notice sometimes the custom Var in chat says 1 even though it should ALWAYS say 2 

### Version

v5.52.0

### What operating system are you using Firebot on?

Windows 10

### Contact Details (Optional)

Badhabitz#8988

PR #1767, opened 2022-05-20
Fix: Random Effect not honoring enabled state

SReject:
Fixes the Random Effect effect not honoring enabled/disabled state.

Ref: #1766


ebiggz (approved):
Noice

ebiggz (changes requested):
We might need to apply this to the Run Sequential Effect too

Issue #1771, opened 2022-05-20
[Bug] $streamer never becomes active according to "Active Chat User" check


CrowbarToolsRobot:
Perry via Discord:

When doing a check like: `$target has role Active Chat User` it fails to see the account as active when targeting the streamer account. Tested both with `Don't allow on active user lists` checked and unchecked on the streamer account.

PR #1774, opened 2022-05-21
fix: Active User Role for Streamer account

brumoen:
### Description of the Change
Using username instead of displayName when chat message is sent internally from Firebot chat field.

### Applicable Issues
#1771 

### Testing
When testing prior to change Streamers with capital letters in their username always fails to show up as an active user in the `Active User Role`, after change has applied testing with a username with capital letters they show up as active users when testing for the `Active User Role`.


ebiggz (commented):
Does this make your username show up as all lowercase in the chat feed? If so, I think we should revert this and instead just do .toLowerCase() when adding users to active/online caches to ensure the keys are always normalized

PR #1775, opened 2022-05-21
fix: Active User Role for streamer account (#1771)

brumoen:
### Description of the Change
Made sure that userDetails has the correct casing for username by applying .toLowerCase().

### Applicable Issues
#1771 

### Testing
Tested to make sure that the streamer shows up with the active user role as well as with the correct casing in viewer list, chat feed.

### Screenshots
Tested with setting for disabling active user disabled:
image
Tested with setting for disabling active user enabled:
image


ebiggz (approved):
Looks good!

Issue #1792, opened 2022-06-14
[Feature Request] Sequential Effect Queue type


iamdadmin:
**Describe the solution you'd like**
Sequential Effect Queue type which is similar to Interval, except it will wait for the currently queued effect to run completely, before triggering the next one, with optional interval to spread them out.

**Additional context**
As discussed in the discord here https://discord.com/channels/372817064034959370/372818514312167424/984718893937487903

I'm raising this issue so I can link it to a merge/pull request for context.

iamdadmin:
Note for any mods - this issue can be closed, the merge was released in 5.52.

PR #1794, opened 2022-06-14
Feat: #1792 adds Sequential Effect Queue type

iamdadmin:
### Description of the Change

- Added a sequential option to effect-queue-runner.js / runQueue() utilising .finally so it doesn't proceed until the effect has finished, with optional delay, if not set or 0 will proceed immediately
- Added to gui in various locations
- Chose icon in the modal for this Effect Queue type as iconClass: "fa-arrow-down-1-9" however if that's not appropriate/not available then another would need to be selected

### Applicable Issues
#1792

### Testing

- Added a new effect queue with new type
- Added a new effect queue of a different type, and edited it to be this effect type
- Edited an effect queue of this type, and changed it to another
- Ran three commands which run a set of effects under the three existing types, and noted non-sequential behaviour
- Ran three commands which run a set of effects under the new Sequential type with 1s interval, and noted sequential behaviour
- Ran three commands which run a set of effects under the new Sequential type with 15s interval, and noted sequential behaviour with 15 seconds delay between the end of the last effect and start of the next effect

I did note the following error while running effects, however this appeared for me regardless of the type of Effect Queue and did not seem to be introduced by my code.

```js
warn: Error while processing effects for queue 523e3430-b7d8-11ec-9873-c1018255e0b7 
{
  message: "Cannot read property 'definition' of undefined",
  stack: "TypeError: Cannot read property 'definition' of undefined\n" +
    '    at validateEffectCanRun (F:\\vscode\\firebot\\Firebot\\backend\\common\\effect-runner.js:53:67)\n' +
    '    at Object.runEffects (F:\\vscode\\firebot\\Firebot\\backend\\common\\effect-runner.js:115:14)'
}
```

### Screenshots

Showing the GUI modal.

Feat #1792 adds Sequential Effect Queue type-img01

Showing the type of Sequential showing in the effect queues grid list.

Feat #1792 adds Sequential Effect Queue type-img02


Firebottle:
Thanks for the PR. This looks good to me, but we'll wait for ebiggz to review as well.

Firebottle (approved):

ebiggz:
Hey @iamdadmin thanks so much for the contribution and sorry it took me so long to take a look. 

I've made a couple tweaks post merge of this PR that I wanted to make you aware of so you aren't caught off guard if you pull latest. 

Your new sequential queue mode is effectively an upgraded version of the Auto mode (displayed as Delay in the UI). It works the same except it can have an optional delay in between the effect lists. So to simplify things for the user, I went ahead and merged your enhancements into the Auto mode and renamed the Auto mode display name from "Delay" to "Sequential" + copied over the icon and description. From the UI perspective, it looks as if the "Delay" mode was simply removed. But it's now using the "auto" type internally. I opted for this as to not break existing setups that are currently using Auto queues, but if you've been running Firebot from the dev environment you may notice some of your queues are broken with this tweak. 

Thanks again for helping make Firebot better!

iamdadmin:
That's awesome - and yes the code was based on auto as you saw, so the change makes sense. Thanks for tagging me with the detailed response @ebiggz - it's appreciated.

Issue #1796, opened 2022-06-15
[Bug] Community gift subs will incorrectly show up as Anonymous


CrowbarToolsRobot:
Dennis.sh via Discord:

This issue was originally reported by CrashKoeck and Nikoheart, We confirmed once again that it's still an issue in today's Nikoheart stream, as his own gifted subs showed as anonymous.

dennisrijsdijk:
To reproduce: gift a non-anonymous community sub from Twitch. **You can not reproduce this without Twitch chat events.**

dennisrijsdijk:
This issue can be closed as #1803 was merged.

zunderscore:
Issues are closed upon a release with the relevant fix/change. The dev team will make sure this is closed at that time.

Nikoheartttv:
Can confirm this change works upon testing within the nightly

CKY-:
Resolved with release v5.53.0

PR #1803, opened 2022-06-25
fix: Anonymous community sub detection in event

dennisrijsdijk:
### Description of the Change

Change the condition on the community gift sub event to properly reflect the status of the sub being anonymous or not.

### Applicable Issues

#1796
Regression from https://github.com/crowbartools/Firebot/pull/1663

### Testing

To reproduce: gift a non-anonymous community sub from Twitch. **You can not reproduce this without Twitch chat events.**

I was unable to test the fix due to testing requiring the usage of real Twitch events, but logic dictates this should fix the issue.

### Screenshots

N/A

### DIFF
--- a/backend/events/twitch-events/gift-sub.js
+++ b/backend/events/twitch-events/gift-sub.js
@@ -42,7 +42,7 @@ exports.triggerSubGift = (subInfo) => {
                             username: gifterDisplayName,
                             subCount: giftReceivers.length,
                             subPlan: subInfo.planName,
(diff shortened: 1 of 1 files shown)


dennisrijsdijk:
As a sidenote, I believe this may make https://github.com/crowbartools/Firebot/pull/1798 obsolete, however I haven't had the time to properly test and investigate.

SReject (approved):

Issue #1807, opened 2022-07-07
[Feature Request] advanced option to ignore backing up "overlay-resources" folder


CKY-:
**Describe the solution you'd like**
I have been heavily using the overlay-resources folder for my rescoreses on my overlays and overlay instances. I keep a separate dev folder on a separate drive. over time  this folder has become maybe 3 gig in size. this is due to images and textures and sound bits and models.  it would be nice to be able to turn off backing this folder up or moveing it out of the root of firebot  all together and asigninging within firebot. so i can point at my dev folder. when i exit firebot it  takes a good  bit maybe 10 min to a half hour to fully shut down because of all the assets that are in that folder. 

due to CORS errors in the overlay scripts all assets have to be within the webserver folder  structure.

a "yes/no" of "should the auto backup include the overlay-resources folder and be allude in manual backups?" option would be nice

**Additional context**

PR #1810, opened 2022-07-09
Feat: include overlay resources in backup option

CKY-:
### Description of the Change
Add an option to include the overlay-resources folder in the backups.

Reason for this is:
I have been heavily using the overlay-resources folder for my rescoreses on my overlays and overlay instances. I keep a separate dev folder on a separate drive. over time this folder has become maybe 3 gig in size. this is due to images and textures and sound bits and models. it would be nice to be able to turn off backing this folder up or moveing it out of the root of firebot all together and asigninging within firebot. so i can point at my dev folder. when i exit firebot it takes a good bit maybe 10 min to a half hour to fully shut down because of all the assets that are in that folder.

due to CORS errors in the overlay scripts all assets have to be within the webserver folder structure.

a "checkbox" with the caption of "Include overlay resources folder" would be added to the backup settings option. 

### Applicable Issues

#1807 

### Testing

With the checkbox checked the resources folder was included in the backup zip
With the checkbox unchecked the resources folder was not included in the backup zip

### Screenshots

image

### DIFF
--- a/backend/backupManager.js
+++ b/backend/backupManager.js
@@ -109,8 +109,15 @@ function startBackup(manualActivation = false, callback) {
     const folderPath = path.resolve(dataAccess.getPathInUserData("/"));
     //archive.directory(folderPath, "profiles");
 
+    let varIgnore = ['overlay-resources/**', 'backups/**', 'clips/**', 'logs/**', 'overlay.html'];
+    const includeResources = settings.backupIncludeResources();
+
+    if (includeResources) {
+        varIgnore = ['backups/**', 'clips/**', 'logs/**', 'overlay.html'];
+    }
+
     archive.glob('**/*', {
-        ignore: ['backups/**', 'clips/**', 'logs/**', 'overlay.html'],
+        ignore: varIgnore,
         cwd: folderPath
     });
 
--- a/backend/common/settings-access.js
+++ b/backend/common/settings-access.js
@@ -146,6 +146,15 @@ settings.backupOnExit = function() {
     return backupOnExit != null ? backupOnExit : true;
 };
 
+settings.backupIncludeResources = function() {
+    const save = getDataFromFile("/settings/backupIncludeResources");
+    return save != null ? save : true;
+};
+
+settings.setBackupIncludeResources = function(backupIncludeResources) {
+    pushDataToFile("/settings/backupIncludeResources", backupIncludeResources === true);
+};
+
 settings.backupBeforeUpdates = function() {
     const backupBeforeUpdates = getDataFromFile("/settings/backupBeforeUpdates");
     return backupBeforeUpdates != null ? backupBeforeUpdates : true;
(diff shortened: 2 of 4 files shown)


backend/backupManager.js
```diff
@@ -109,8 +109,15 @@ function startBackup(manualActivation = false, callback) {
     const folderPath = path.resolve(dataAccess.getPathInUserData("/"));
     //archive.directory(folderPath, "profiles");
 
+    let varIgnore = ['overlay-resources/**', 'backups/**', 'clips/**', 'logs/**', 'overlay.html'];
+    const includeResources = settings.backupIncludeResources();
+
+    if (includeResources) {
+        varIgnore = ['backups/**', 'clips/**', 'logs/**', 'overlay.html'];
```
Firebottle, line 116: Should we be including backups inside of the backups? Maybe backups are the one thing that should always be ignored.

backend/common/settings-access.js
```diff
@@ -146,6 +146,15 @@ settings.backupOnExit = function() {
     return backupOnExit != null ? backupOnExit : true;
 };
 
+settings.backupIncludeResources = function() {
+    const save = getDataFromFile("/settings/backupIncludeResources");
+    return save != null ? save : true;
+};
+
+settings.setBackupIncludeResources = function(backupIncludeResources) {
+    pushDataToFile("/settings/backupIncludeResources", backupIncludeResources === true);
```
Firebottle, line 155: We'll probably want to default to off so that current functionality isn't changed.

Firebottle (changes requested):
Looks good, and this will be a good thing to have. Only things I see are that maybe backups should always be ignored so the backup size doesn't compound on itself. Also, we should default this to off so that current live functionality still works the same.

backend/common/settings-access.js
CKY-: that is how it currently works.  you must uncheck the "Include Resources" to turn it off as the current functionality is to include them.

CKY-: > We'll probably want to default to off so that current functionality isn't changed.

the verbage here  would you prefer include or ignore?

backend/backupManager.js
CKY-: how verbrous would you like the variable name?

rather than checking if this is true depending on the the verbiage of include vs ignore 
i could invert or check if ignore is false
```js
let varFilesAndFildersToIgnore = ['backups/**', 'clips/**', 'logs/**', 'overlay.html'];
const includeResources = settings.backupIncludeResources();

 if (!includeResources) {
        varFilesAndFildersToIgnore = [' overlay-resources/**','backups/**', 'clips/**', 'logs/**', 'overlay.html'];
```

```js
let varIgnore = ['backups/**', 'clips/**', 'logs/**', 'overlay.html'];
const includeResources = settings.backupIncludeResources();

 if (includeResources == false) {
        varIgnore = [' overlay-resources/**','backups/**', 'clips/**', 'logs/**', 'overlay.html'];
```

Issue #1833, opened 2022-07-28
[Feature Request] For Play Video Effect add volume slider for Twitch Clip and Random Twitch Clip


DragynsLair:
For Play Video Effect add volume slider for Twitch Clip and Random Twitch Clip

PR #1841, opened 2022-08-02
feat: add volume support for Twitch clips

zunderscore:
### Description of the Change
Adds volume support to Twitch clips played in overlay. Also rewrites Twitch clips in the overlay to play the raw video URL in a `<video>` element instead of a Twitch player embedded in an `<iframe>`.

**NOTE:** Users may have to adjust volume settings for any **Play Video** effects they have that are set to either **Twitch Clip** or **Random Twitch Clip**, as they will now play at the effect's default volume of 50%, whereas before they were all playing at 100%.

### Applicable Issues
#1833

### Testing
Verified that both specific and random Twitch clips play in the overlay at the specified volume.

### Screenshots
image

### DIFF
--- a/backend/effects/builtin/clips.js
+++ b/backend/effects/builtin/clips.js
@@ -180,9 +180,7 @@ const clip = {
         const { effect } = event;
         const clip = await clipProcessor.createClip(effect);
         if (clip != null) {
-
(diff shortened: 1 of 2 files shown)


ebiggz (approved):

ebiggz:
bad ass

PR #1862, opened 2022-08-12
feat: make the URL allowlist discern on the path component of a URL

AliciaBytes:
**Issue:**
Currently the bot's allowlist for URL moderation doesn't work for the path component of a domain. I'd like to also filter not just on the domain, but also path part. For example for multistreams I'd love to allow the urls to the channels of my costreamers only, but not all twitch links in general.

For example allowing just `https://twitch.tv/good_streamer` won't work since the current regex only picks up the `https://twitch.tv/` part and everything after gets ignored.

image

leaving me the choice to whitelist twitch as a whole or not at all.

**Fix:**

Changing

```js
const getUrlRegex = () => {
    return /\b(?:https?:(?:\/\/)?)?(?:[a-z\d](?:[a-z\d-]{0,253}[a-z\d])?\.)+[a-z][a-z\d-]{0,60}[a-z\d](?:$|[\\/]|\w?)/gi;
};
```

and adding a single `+` to the last group will make the path component part of the url and allow discerning on it for the allowlist:

```js
const getUrlRegex = () => {
    return /\b(?:https?:(?:\/\/)?)?(?:[a-z\d](?:[a-z\d-]{0,253}[a-z\d])?\.)+[a-z][a-z\d-]{0,60}[a-z\d](?:$|[\\/]|\w?)+/gi;
};
```

image

### DIFF
--- a/src/backend/utility.js
+++ b/src/backend/utility.js
@@ -17,7 +17,7 @@ const escapeRegExp = (str) => {
 };
 
 const getUrlRegex = () => {
-    return /\b(?:https?:(?:\/\/)?)?(?:[a-z\d](?:[a-z\d-]{0,253}[a-z\d])?\.)+[a-z][a-z\d-]{0,60}[a-z\d](?:$|[\\/]|\w?)/gi;
+    return /\b(?:https?:(?:\/\/)?)?(?:[a-z\d](?:[a-z\d-]{0,253}[a-z\d])?\.)+[a-z][a-z\d-]{0,60}[a-z\d](?:$|[\\/]|\w?)+/gi;
 };
 
 /**


zunderscore:
Hey @AliciaBytes, thanks for the PR! Can you please rebase off of and target the `v5` branch instead of the `master` branch? We do all active development on the specific version branch. Thanks!

AliciaBytes:
@zunderscore thanks for letting me know, just did that.

AliciaBytes:
Is there anything else I can do to help move this along?

ebiggz:
@AliciaBytes Apologies for the delay! This is great - thanks for taking the time to improve this regex.

ebiggz (approved):

Issue #1866, opened 2022-08-17
[Bug] Firebot Polly doesn't conform with AWS Polly SSML standards


dennisrijsdijk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Some TTS messages, like ones containing < (without an ssml tag), & (unencoded) or some ssml tags like amazon:effect (due to missing `<speak>MESSAGE</speak>`) result in an error 400 from aws with invalid ssml, which means the message doesn't get spoken out.

EDIT 8/25/2022: After #1867 was merged, the only remaining issue is unencoded starting brackets (`<`) that do not belong to an ssml tag.

### Expected Behavior

SSML Messages should be encoded according to [the AWS Polly instructions](https://docs.aws.amazon.com/polly/latest/dg/escapees.html)
When this is the case, messages will properly be spoken out.

### Steps To Reproduce

1. Set up a Text-To-Speech (Amazon Polly) effect with SSML enabled
2. In the text field, add `some text < some more text`, `some text & some more text` or `<amazon:effect name="whispered">some whispered text</amazon:effect>` and run the effect
3. Now to verify a working example, try `some text &lt; some more text`, `some text &amp; some more text` or `<speak><amazon:effect name="whispered">some whispered text</amazon:effect></speak>`

`<speak></speak>` should be present on any tts message that has SSML enabled
any ampersand (&) should be escaped to &amp; as per [the AWS Polly instructions](https://docs.aws.amazon.com/polly/latest/dg/escapees.html)
single and double quotes don't need to be escaped according to testing.
< and > should only be escaped if they're not a part of an ssml tag. If they're escaped as part of an SSML tag, the SSML tag won't work.

### Version

5.53.2

### What operating system are you using Firebot on?

Windows 11

### Relevant log output

```shell
N/A
```

### Contact Details (Optional)

Dennis.sh#9272

PR #1867, opened 2022-08-17
fix: resolve some SSML issues with Amazon Polly

dennisrijsdijk:
### Description of the Change

Introduce 2 fixes for Amazon Polly SSML.
Introduce `<speak></speak>` when SSML is selected, fixes some SSML tags not working.
Escape `&` character to make it work in an SSML message.
Note: a fix for the `<` character is NOT included, more work needs to be put into filtering so genuine SSML tags aren't escaped/encoded

### Applicable Issues

#1866 

### Testing

When testing the fix, every example text in the reproduction steps, except for the one with a `<` character now works. See the note above for clarification.

### Screenshots

N/A

Thanks to @Nikoheartttv (Nikoheart#0001) for helping with this issue.


ebiggz (approved):
Thanks for this PR, @dennisrijsdijk !

Issue #1871, opened 2022-08-19
[Bug] Event Filter "Chest Purchase" not changing its Check Value


Nashuja2022:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If you want to the filter inside the "Chest Purchase" Event to "NOT A Gift", Firebot won't save the filter as "NOT A Gift", but keeping it as "A Gift". This way, the "NOT A Gift"-path won't be triggered either.

### Expected Behavior

Either: Have the option "NOT A Gift" to be saved by Firebot, OR
Be able to select a "is no" instead of only a "is" comparator for the Event, so you can create a "is not" "A Gift" Filter, instead of a "is" "NOT A Gift". 

### Steps To Reproduce

1. Add a "Chest Purchase" Event to Firebot
2. Create a Filter
3. Chance the Filter to Chest Purchase and the Value to check to "NOT A Gift"
4. Save the Settings
5. Look at the saved filter

### Version

5.53.2

### What operating system are you using Firebot on?

Windows 10

### Contact Details (Optional)

Nashuja#4453

PR #1873, opened 2022-08-19
fix: streamloots chest is a gift and is not a gift streamloots chest event #1872

CKY-:
### Description of the Change
Fixed the logic for the the selection of the "A Gift" "Not A Gift" on the streamloots chest event 

### Applicable Issues

#1871

### Testing

Tested using the simulate event to create chest purchase of a gift and non gift chest 
this is now working as intended.


ebiggz (approved):
Thanks CKY!

Issue #1797, opened 2022-06-18
[Feature Request] Add time limit for TTS effect


mgway:
**Describe the solution you'd like**
It would be great if you could limit TTS to the first Y seconds. There's only so many times one can hear TTS from a message full of 7s that results in Microsoft Sam spelling out the number. 

**Additional context**
I'm aware that there is a `Clear Effects` effect, but it would be nice to truncate the sound without intervention.

zunderscore:
The `$textSubstring` variable can accomplish this today. For example: if you have a channel reward that triggers TTS based on the reward message entered by the viewer, instead of using the `$rewardMessage` variable on its own and having TTS read the entire message, you could use `$textSubstring[$rewardMessage, 0, 100]` to only read the first 100 characters.

mgway:
Thanks for the tip. I'll rewrite this issue for adding a time limit only since there is still the possibility of relatively short messages producing unfortunately long TTS results.

PR #1874, opened 2022-08-24
#1797 Add time limit to Polly and native TTS

mgway:
### Description of the Change
Added a maximum duration parameter to help curb overly long winded Polly and native TTS messages. 

### Applicable Issues
https://github.com/crowbartools/Firebot/issues/1797

### Testing
- Added new chat !command trigger
- Added Polly TTS effect
  - Set text to the Shungite copypasta (it takes about 30 seconds for built-in TTS to read)
- Did not set new maximum duration property
- Tested event. Observed that the whole message was read out
- Edited event and set maximum duration to 7 seconds
- Tested event. Observed that only the first 7 seconds were read, approximately one sentence
- Edited event again and set maximum duration to -1
- Tested event. Observed that the whole message was read out

Repeated the above tests but with the built-in text to speech effect instead of Polly

### Screenshots
image

### DIFF
--- a/src/backend/integrations/builtin/aws/text-to-speech-polly-effect.js
+++ b/src/backend/integrations/builtin/aws/text-to-speech-polly-effect.js
@@ -146,6 +146,13 @@ const playSound = {
             </div>
         </eos-container>
 
+        <eos-container header="Maximum Duration" pad-top="true">
+            <div class="input-group">
+                <span class="input-group-addon" id="delay-length-effect-type">Seconds</span>
+                <input ng-model="effect.maxSoundLength" type="text" class="form-control" aria-describedby="delay-length-effect-type" type="text" replace-variables="number">
+            </div>
+        </eos-container>
+
         <eos-container header="Sound" pad-top="true">
             <label class="control-fb control--checkbox"> Wait for sound to finish <tooltip text="'Wait for the sound to finish before letting the next effect play.'"></tooltip>
                 <input type="checkbox" ng-model="effect.waitForSound">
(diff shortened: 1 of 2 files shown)


src/gui/app/services/tts.service.js
```diff
@@ -92,11 +92,19 @@
                 msg.lang = 'en-US';
 
                 speechSynthesis.speak(msg);
+
+                const intMaxSoundLength = parseInt(maxSoundLength);
+                if (intMaxSoundLength > 0) {
+                    setTimeout(function() {
+                        speechSynthesis.cancel()
+                    }, maxSoundLength * 1000)
```
ebiggz, line 100: If i'm reading this right, it looks like this timeout always fires for each tts that comes through that has a max sound length defined. Given that `speechSynthesis.cancel()` clears the current tts and the tts queue, this means we nuke the queue for each tts even if they dont actually surpass the length limit.

The solution for this is to clear the timeout when the given tts ends via [`SpeechSynthesisUtterance`'s "end" event](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisUtterance/end_event). You should be able to do that like so:
```ts
const timeoutId = setTimeout(speechSynthesis.cancel, maxSoundLength * 1000);
msg.onend = () => {
   clearTimeout(timeoutId);
}
```
> Note: I've also simplified setTimeout by passing a reference to the speechSynthesis.cancel function instead of wrapping it in a new function.

ebiggz (changes requested):
Thanks for this PR! Looks great - I believe theres just a small bug that needs looking at and then it should be good to go.

src/gui/app/services/tts.service.js
mgway: Ah, I didn't realize that all native TTS messages were handed off to the speech processor immediately and it managed the queue. Since there's no way to extract utterance boundaries or duration, I've removed the native TTS portion of this change. The only way that I could see that working would be to add a `Wait for the sound to finish before letting the next effect play`-like option to the native TTS option and turning the `read-tts` event into an async one but that seemed like a bit too much heavy lifting for what looked like a small change.

ebiggz (approved):

PR #1937, opened 2022-09-14
chore: refactor third party emote caching

zunderscore:
### Description of the Change
Refactored third party emote loading in TypeScript

### Applicable Issues
N/A

### Testing
Verified third party emotes still load

### Screenshots
N/A


ebiggz (changes requested):
I really dig the direction you took here. 

If we are getting our hands dirty with this, we could take it one step further to get the code super dry. We can make `ThirdPartyEmoteProvider` an abstract class instead of an interface, allowing us to define a default implementation of `getAllEmotes()`.

Here's a gist where I converted `ThirdPartyEmoteProvider` to an abstract class and ported `ffz.ts` to use is as a reference if you like this route:
https://gist.github.com/ebiggz/31b539754cfa8b3dd9b137d014f76bde

ebiggz (approved):

Issue #1940, opened 2022-09-16
[Bug] "Cooldown command" effect broken


sheuronazxe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

![image](https://user-images.githubusercontent.com/9971945/190640181-aec7b4b4-65f3-4e6f-8fed-b7950edd502d.png)

Try to add effect show red box "Please select a command or tag" but none command can be selected.

### Version

5.54.2

### What operating system are you using Firebot on?

Windows 11

zunderscore:
This appears to be an environmental issue. Please create a thread in the **issues** channel of our Discord and we'll be happy to help troubleshoot.

CKY-:
Adding a Command Tag to commands Enables this. This should work independently of tags.

CKY-:
https://github.com/crowbartools/Firebot/blob/11045c290f35d0fab9156278d0ea205b9d0dbc03/src/backend/effects/builtin/cooldown-command.js#L134

this seems to be the issue

zunderscore:
Fixed in v5.55.0

PR #1983, opened 2022-11-22
fix: cooldown command effect  #1940

CKY-:
### Description of the Change

Allow for Single Command to be displayed Indigent of Command Tags,
Set Single Command as the default selection if Commands exist.  

### Applicable Issues

#1940

### Testing

Without Command Tags 

image

with Command Tags 
image


ebiggz (approved):
Thanks for this CKY!

Issue #1608, opened 2022-03-03
[Feature Request] Event trigger for time of day


zunderscore:
**Describe the solution you'd like**
I'd love to see event triggers that occur based on time of day. For example, if I want an event to trigger every Friday at 1PM local time, I could set a list of effects of my choosing to run at that time.

zunderscore:
After a very long time, I'm so happy to finally be able to say...

Added in v5.55.0 ❤️

PR #2004, opened 2022-12-27
feat: Scheduled tasks

zunderscore:
### Description of the Change
Adds the new **Scheduled Task** trigger type as a subset of Timers. Scheduled tasks are scheduled like cron jobs, running in a similar context as timers.

### Applicable Issues
#1608

### Testing
Validated all UI, created items with both simple and advanced schedules and verified they all ran on time as intended.

### Screenshots
image
image
image


ebiggz (commented):
Really dig this feature and appreciate the work involved getting it implemented so cleanly. Thank you so much!

The only thing that gives me pause is the name. Its' a difficult feature to concisely name to be sure. But something about the word "task" feels out of place in the context of Firebot. And the "Timers / Scheduled Task" tab name is unsightly in the header (just because of the length). 

As for "task", personally I think something like "Scheduled Effects" would fit in better and make more sense. 

As for the tab name, I wonder if perhaps generalizing it as "Time-based" would be easier on the eyes.

Thoughts? I don't want to be a stick in the mud for something so mundane as a name. So if you feel strongly that my suggestions don't improve anything, feel free to consider this approved. At the very least, I'd love to see the tab name addressed, even if we decide to continue with "task".

zunderscore:
Yeah, I was trying to figure out a better way to name the tab but came up pretty empty. I really like "Time-Based" for the tab, then yeah, I'm down for "Scheduled Effects", but "Tasks" just felt more familiar from how those features are typically presented in other places, like OS tasks, etc. I'll probably still call it `scheduledTasks` under the hood unless you have objections.

ebiggz:
@zunderscore Totally understand the perspective of "tasks", definitely terminology I'd expect to see in things like that. My main thing with it in Firebot is the word effectively represent  "Effects" or an "Effect list" in this case. And I just get weary about introducing new terminology when talking to users (perhaps overly so) so I tend to advocate for being as consistent as possible. 

I'd totally be down if we wanted to consider Task as the official term for list of actions (effects) in v6. I think its better than "Effect List" or "Action List". Reminds me of iOSs shortcuts app which has Actions and Shortcuts (list of actions). Main thing is I just want to be consistent with whatever we chose. 

And yes, if you are cool with using "Scheduled Effects" I am totally fine with just changing the customer facing text. No need to refactor all the code to match.

ebiggz (approved):

Issue #2012, opened 2023-01-06
[Bug] system sub commands is active is and auto delete trigger are not hooked to anything. 


CKY-:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When unchecking IS Active check box the command disables but is still allowed to run 
  
![image](https://user-images.githubusercontent.com/8982158/211092809-99a6f73b-6f06-4b27-9dda-10f16b653fef.png)
![image](https://user-images.githubusercontent.com/8982158/211092993-c7c92669-c4fc-4d54-ac90-6d66c35d45d0.png)

### Expected Behavior

not to elevate the command when it has been deactivated 

### Steps To Reproduce

go to system commands 
expand quotes
expand add 
uncheck is active and save 
send in chat the command to add a quote 

### Version

v5.54.3

### What operating system are you using Firebot on?

Windows 10

CKY-:
this is not limited to add it seems all subcommands are affected

zunderscore:
Fixed in v5.55.0

PR #2021, opened 2023-01-13
Fix: 2012 An attempt to fix sub command activation

CKY-:
### Description of the Change

This is to try to fix the "Is Active" of bug #2012 
I have been struggling with a better solution that takes in account both "system" and "custom" sub command activation.  

### Applicable Issues

#2012 

### Testing

i have ran this for a week with no adverse effects but i'm sure something is not correct and can not for the life me see what? 
i would like a code review and am expecting to have to change it.


ebiggz (approved):

ebiggz:
Thanks for this CKY

PR #2023, opened 2023-01-14
feat: raw OBS WebSocket request effect

zunderscore:
### Description of the Change
Adds a new **Send Raw OBS WebSocket Request** effect, similar to the **HTTP Request** effect.

### Applicable Issues
N/A

### Testing
Verified effect works

### Screenshots
image
image


ebiggz (changes requested):
Awesome idea! My only suggestion is to remove the option to put response in a variable and instead utilize the new effect output system (this is one of the scenarios the output system is intended to help with. HTTP effect still needs to be updated to it)

ebiggz (approved):

PR #2026, opened 2023-01-14
feat: allow effect outputs to be bubbled to parent lists, update http request effect to use outputs

ebiggz:
### Description of the Change

- Adds option to have effect outputs be applied to parent lists (run effect list, random effect, and conditional effects)
- Update HTTP Request effect to use outputs


zunderscore (approved):
Looks good to me :)

Issue #2032, opened 2023-01-16
[Feature Request] Linux OBS Text Sources


dennisrijsdijk:
OBS On Linux uses the `Text (FreeType 2)` source with inputKind `text_ft2_source_v2` rather than GDI+

**Describe the solution you'd like**
The **Set OBS Source Text** effect should be updated to include support for FreeType 2 text sources.

**Additional context**
The settings object of the FreeType 2 text source is slightly different from GDI+. I have included it below.
```json
"inputSettings": {
    "from_file": true,
    "text": "some text",
    "text_file": "/home/dennis/Firebot/license.txt"
  }
  ```

zunderscore:
Added in v5.55.0

PR #2033, opened 2023-01-16
feat: Linux OBS text source update

zunderscore:
### Description of the Change
Adds Linux support for the **Set OBS Source Text** effect.

### Applicable Issues
#2032

### Testing
Verified this still works on Windows. Linux testing to be completed by @dennisrijsdijk.

### Screenshots
N/A

### DIFF
--- a/src/backend/integrations/builtin/obs/obs-remote.ts
+++ b/src/backend/integrations/builtin/obs/obs-remote.ts
@@ -335,19 +335,34 @@ export async function setSourceMuted(sourceName: string, muted: boolean) {
 
 export async function getTextSources(): Promise<Array<OBSSource>> {
   const sources = await getAllSources();
-  return sources.filter((s) => s.typeId === "text_gdiplus_v2");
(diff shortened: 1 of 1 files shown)


dennisrijsdijk:
Hm, looks like I'm getting an Unhandled promise rejection on Linux.
```js
{
  message: "Cannot read property 'filter' of null",
  stack: "TypeError: Cannot read property 'filter' of null\n" +
    'at Object.getTextSources [as callback] (/home/dennis/Code/Firebot/build/backend/integrations/builtin/obs/obs-remote.js:282:20)'
}
```

I'll investigate and see if I can find the cause, but I may not be able to get to the cause today

dennisrijsdijk:
Apparently, connected is false for me, which returns null on `getAllSources`. I'll try a fresh profile with fresh OBS credentials, but I feel like something's not right here

dennisrijsdijk:
Looks like that was weirdness on that particular profile. It works perfectly, letting me set text, set file, and go back to text.

zunderscore:
Awesome! Thanks for testing :)

Issue #2036, opened 2023-01-20
[Feature Request] MetaData property path to accept variable data


CKY-:
**Describe the solution you'd like**
please make property paths in set metadata to show variable inputs 

**Additional context**

CKY-:
Added in v5.55.0

PR #2040, opened 2023-01-21
Feat: metadata variables in property path

CKY-:
### Description of the Change

adds variables to property paths of set metadata

### Applicable Issues

#2036

### Testing

this was tested and working in preset effects using the preset variable as the path of what to change.  

### Screenshots

image


zunderscore:
I love adding the vars button to that field, but let's keep the UI here consistent and keep the position inside the input box, since it's a full-width field.

zunderscore (approved):
Looks great!

PR #2049, opened 2023-02-01
fix: Enable use-text-area on relevant textboxes

dennisrijsdijk:
### Description of the Change

This PR enables larger multi-line textboxes on the **Text** and **Replacement Text** inputs for the **Write To File** effect, as well as the input for **Set OBS Source Text**

### Applicable Issues

N/A

### Testing

I've tested and verified that text is still properly written to file (with multiline/newline support) and that previously created effects still work properly.
For the OBS Source Text, I've tested that text properly gets set on the text source (with multiline/newline support) and that previously created effects still work properly.

### Screenshots

N/A


zunderscore (approved):
Looks great!

Issue #1955, opened 2022-10-04
[Feature Request] Add the ability to save the replay buffer in OBS


dadthegamer:
**Describe the solution you'd like**
Add an effect under the OBS category to save the replay buffer.

PR #2050, opened 2023-02-01
feat: OBS Save Replay Buffer effect (#1955)

zunderscore:
### Description of the Change
Adds new Save OBS Replay Buffer effect

### Applicable Issues
#1955

### Testing
Verified replay buffer saved when enabled and active in OBS

### Screenshots
image
image


ebiggz (approved):
nooice

Issue #2058, opened 2023-02-04
[Bug] Resource Token for Play Video Effect expires before the end of long videos.


dennisrijsdijk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When the video duration isn't specified, the resource token for videos longer than 20 seconds (it seems that chrome and chromium preload around 20 seconds) will expire before the whole video can be preloaded/cached, resulting in the video ending early with an error.

### Expected Behavior

Videos should finish playing, regardless of the length, filesize etc.

### Steps To Reproduce

1. Create a play video effect with a long local video (> 20 secs, example video to use is https://d3g7htsbjjywiv.cloudfront.net/assets/common/images/media-page/redesigned-footage/footage-sizzle.webm)
2. Do not specify a duration.
3. Open the overlay in the browser and **click on the page to enable autoplay**
4. Open devtools and navigate to the console.
5. Play the video on overlay.
6. Notice that when the video ends prematurely, an error appears in console.

### Version

5.55.2

### What operating system are you using Firebot on?

Windows 11

### Relevant log output

```shell
GET http://localhost:7472/resource/5249smu9po8u39x60bl1tk 404 (Not Found)
```

PR #2059, opened 2023-02-04
fix: long videos resourceToken expires before video finish

dennisrijsdijk:
### Description of the Change

Resource token length for videos is now dynamically fetched from the overlay instead of relying on being set as video duration by the user.

### Applicable Issues

#2058 

### Testing

I've tested to make sure that there's no noticeable delay on video start, and the video would complete fully.

### Screenshots

N/A


ebiggz (approved):
Looks good!

PR #2060, opened 2023-02-05
feat: Discord Timestamps Variable

Firebottle:
### Description of the Change
This adds in a new variable that generates discord timestamps. This is meant to really only ever be used with the "send discord message" effect.  This will generate text similar to "<t:1543392060:R>". It supports all of the formats that discord supports. I've also tried to make it flexible as possible so that people can use it to announce stream times and other things.

- Can exclude the date to use the current date at the specified time.
- Can use 'now' instead of a time to use the current date and time. This allows people to use the current time in whatever format they want.
- Can use any discord timestamp format.

https://discord.com/developers/docs/reference#message-formatting-timestamp-styles

### Applicable Issues
- No issues, but it is important to note that discord will change the output format based on if the end user has 12 hour or 24 hour time set. I've tried to include both in the examples. This doesn't change how the user enters time into firebot.

### Testing
-  $discordTimestamp[2076-01-26 13:00:00]
-  $discordTimestamp[13:00:00]
-  $discordTimestamp[2076-01-26 13:00:00, R]
-  $discordTimestamp[13:00:00, R]
-  $discordTimestamp[now, R]
-  $discordTimestamp[]
-  $discordTimestamp[blahblahblah] (will output [Incorrect date format])
- $discordTimestamp[now, LOL] (will fallback to discord default format)

### Screenshots
None

### DIFF
--- a/src/backend/variables/builtin-variable-loader.js
+++ b/src/backend/variables/builtin-variable-loader.js
@@ -56,6 +56,7 @@ exports.loadReplaceVariables = () => {
         'custom-variable-expired-data',
         'custom-variable-expired-name',
         'date',
+        'discord-timestamp',
         'donation-amount-formatted',
         'donation-amount',
         'donation-from',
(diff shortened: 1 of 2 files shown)


ebiggz (approved):
this is awesome, didnt even know about this discord feature!

Issue #1978, opened 2022-11-07
[Feature Request] Known Bots Role


SkullGaming31:
**Describe the solution you'd like**
could you make it so the known bots role automatically disables all the stats, so they cant gain currency or watch time in a channel or even give us an option to disable/enable this feature.

PR #2061, opened 2023-02-05
feat: Auto flag bots to stop stat generation (#1978)

Firebottle:
### Description of the Change
Implements #1978

This adds a setting to the database settings page to auto flag bots. If this is on, then when a user goes online that is a known bot it will automatically set their "disable stat generation" and "remove from active viewers list" settings to on. 

### Applicable Issues
- I have defaulted this to ON for all users, as I think that is what 99% of users will want.
- This flags users as if the streamer had selected the "disable stat generation" option in the user modal. This means once the option is set for the bot, then there is no un-doing it other than manually unselecting it. Turning the auto flag setting off will not retroactively unflag all of the bots.
- If a user unflags a bot manually, this will auto flag them again once the bot "goes online". I don't know of a way to tell if a setting has been manually updated or not previously in order to prevent this from happening.

### Testing
- Remove "disable stat generation" flag from streamer. Then reconnect to chat. Confirm all known bots have the two options on, and confirm the streamer account still has the option off. This should confirm that it is adding the flag to bots, but not regular users.

### Screenshots
None


ebiggz (approved):
All of this looks great!

PR #2066, opened 2023-02-07
feat: OBS Source Screenshot

dennisrijsdijk:
### Description of the Change

This Pull Request adds an effect that takes a screenshot of an OBS Source or Scene, saving it to a file.
It exposes OBS Websocket's quality and size scaling settings as optional options.
Internally, it uses the `GetSourceScreenshot` call rather than `SaveSourceScreenshot` to allow saving the image on the firebot machine in a multi-PC setup.

The OBS Source dropdown is searchable both by source/scene name, as well as type (currently the id of the type rather than the human friendly name, as OBS Websocket doesn't expose that.)

### Applicable Issues

N/A

### Testing

I have verified that the image file gets written properly when OBS is connected and sends back the image as a base64 URI. I have also verified that `null` is returned when OBS is not connected or when there is an error, and there is no attempt to save the file.

### Screenshots

image
image
image
image
image
image


ebiggz (approved):

ebiggz:
Great stuff, thanks for the contribution!

Issue #1819, opened 2022-07-13
[Bug] Multi-month subs don't show an alert for subsequent subs (month 2, month 3 etc.)


dennisrijsdijk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a user has been gifted a multi-month sub, and shares a month from that gift sub, Firebot doesn't fire any of its sub alerts (all of the events are used, and work normally with regular resubs, gift subs etc.)

Below is an example of the missing sub event.

![image](https://user-images.githubusercontent.com/70665154/178719286-cad60649-7b13-4765-95ee-46c347d8787c.png)

### Expected Behavior

Multi-month sub shares should trigger an event.

### Steps To Reproduce

1. Ensure all sub related events are set up
2. Gift a multi-month sub
3. Wait a month
4. Share the sub on Twitch
5. Notice no event trigger

### Version

5.53

### What operating system are you using Firebot on?

Windows 10

### Relevant log output

```
01:44:58 GMT+0100 (British Summer Time) - debug: [v5.52.0] Received streamelements event: 
{
  _id: '62ad2009ea23a40b0a9dfb15',
  channel: '5e57d6130335bf8290b2f146',
  type: 'subscriber',
  provider: 'twitch',
  flagged: false,
  data: {
    username: 'funkadelichika',
    providerId: '145142816',
    displayName: 'funkadelichika',
    amount: 7,
    tier: '1000',
    message: 'Thanks to @Datenshi1325 for my sub gift! Butts.',
    quantity: 0,
    avatar: 'https://static-cdn.jtvnw.net/jtv_user_pictures/d91c2142-1507-4f5d-a639-058436558aca-profile_image-300x300.png'
  },
  createdAt: '2022-06-18T00:44:57.110Z',
  updatedAt: '2022-06-18T00:44:57.110Z',
  activityId: '62ad2009ea23a40b0a9dfb15'
}
```

### Contact Details (Optional)

Dennis.sh#9272

zunderscore:
Resolved in v5.59.0

PR #2258, opened 2023-10-10
Fix multi-month gifted subscriptions not triggering the sub event.

dennisrijsdijk:
### Description of the Change

Fixes the Sub event not firing when a user resubs from a multi-month gifted subscription.

### Applicable Issues

#1819 

### Testing

This change has been tested during a stream with several regular resubs and a multi-month resub.
Regular resubs happened as normal, and the multi-month gifted resub also fired the Sub event.


dennisrijsdijk:
Superseded by #2269

Issue #2287, opened 2023-12-13
[Bug] Inconsistencies with "Wait for sound to finish"


dennisrijsdijk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Repeatedly playing a sound with "wait for sound to finish" can sometimes cause the duration check to lock up. When this is used in effect queues or effects lists, it can cause a deadlock to the queue or list.

### Expected Behavior

Playing a sound with "wait for sound to finish" should always result in the effect waiting the appropriate duration, then continuing execution.

### Steps To Reproduce

1. Create a new command
2. Add a "Play Sound" effect
3. Select a short sound (couple seconds)
4. Enable "Wait for sound to finish"
5. Save the effect
6. Add a "Log Message" effect, set it to debug and add any message.
7. Open Developer Tools
8. Run the command several times. At some point, it should stop producing log entries.

### Version

5.59.1

### What operating system are you using Firebot on?

Linux Distro

### Contact Details (Optional)

dennisoti

PR #2288, opened 2023-12-13
Fix Howler issue related to caching for getSoundDuration (#2287)

dennisrijsdijk:
### Description of the Change

This PR moves the load and loaderror event listeners into the Howl constructor, mitigating audio caching issues that cause the sound to load before we can subscribe to the event listener.

### Applicable Issues

#2287

### Testing

Running the same test on both builds, I confirm the issue only happens pre-changes.

### DIFF
--- a/src/gui/app/services/sound.service.js
+++ b/src/gui/app/services/sound.service.js
@@ -159,17 +159,14 @@
 
                     const sound = new Howl({
                         src: [path],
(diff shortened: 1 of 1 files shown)


dennisrijsdijk:
converted to draft, other testing highlighted more issues

ebiggz (approved):

Issue #2273, opened 2023-11-22
[Bug] Effect outputs don't bubble DOWN into lists


Lordmau5:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Line 181 of effect-runner.js checks if the passed argument has a property called `outputs` and if it doesn't, it sets it to an empty object.
https://github.com/crowbartools/Firebot/blob/1b77e38a1eaeb7fa59581e2017bc9b0bd1646e64/src/backend/common/effect-runner.js#L181-L183

This is however **not** passed to it when called through an effect list
https://github.com/crowbartools/Firebot/blob/1b77e38a1eaeb7fa59581e2017bc9b0bd1646e64/src/backend/effects/builtin/effect-group.js#L154-L157
https://github.com/crowbartools/Firebot/blob/1b77e38a1eaeb7fa59581e2017bc9b0bd1646e64/src/backend/effects/builtin/effect-group.js#L165-L168

These objects are missing the outputs from that effect.

An example setup to replicate this issue can be found here:
[Bubble Down Effect Outputs Test.zip](https://github.com/crowbartools/Firebot/files/13443886/Bubble.Down.Effect.Outputs.Test.zip)

### Expected Behavior

For effect outputs to bubble down into effect lists

### Steps To Reproduce

1. Download the setup archive
2. Extract it
3. Import it into Firebot
4. Test the command

### Version

5.58.1

### What operating system are you using Firebot on?

Windows 11

### Relevant log output

```shell
2023-11-22 19:50:36.0024 - info: [v5.58.1-nightly-23.09.30] This will output correctly: <!DOCTYPE html>

<html lang="en">
<head>
    <meta charset="utf-8" />
    <title>It's not working! Should I blame caching?</title>
    <link href="https://fonts.googleapis.com/css?family=Roboto&amp;text=%20AYabcdgehilmnostuvy1" rel="stylesheet" type="text/css">
    <style>
        body { font-family: 'Roboto', sans-serif; -webkit-text-stroke-width: 0.1px; }
        section { margin: 10em auto 0px auto; width: 15em; }
        h1 { color: #404040; font-size: 10em; line-height: 0.8em; margin: 0; padding: 0; text-align: center; }
        h1 sup { color: #C0C0C0; font-size: 5%; font-weight: normal; line-height: normal; vertical-align: top; }
        aside { color: #C0C0C0; font-size: 0.6em; margin-top: 1em; }
[...]

PR #2317, opened 2023-12-31
fix(effects): effect outputs now bubble down into nested effect lists (#2273)

zunderscore:
### Description of the Change
Effect outputs now bubble down into nested effect lists, including the **Run Effect List**, **Loop Effects**, **Run Random Effect**, and **Run Sequential Effect** effects.

### Applicable Issues
#2273

### Testing
Verified effect outputs bubbled down into a nested effect list via **Run Effect List** and that loops in **Loop Effects** persisted outputs between runs

### Screenshots
N/A

### DIFF
--- a/src/backend/common/effect-runner.js
+++ b/src/backend/common/effect-runner.js
@@ -74,7 +74,7 @@ function validateEffectCanRun(effectId, triggerType) {
 
 }
 
-async function triggerEffect(effect, trigger) {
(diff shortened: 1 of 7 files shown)


dennisrijsdijk:
Has this been double checked for random effects, sequential effects?

ebiggz (approved):

PR #2324, opened 2024-01-03
fix(misc): fixed several UI/log message/comment typos

CKY-:
### Description of the Change

fixes some typos and misspellings in the ui

### Testing

loaded firebot and confirmed that the changes where made.

### DIFF
--- a/src/backend/twitch-api/resource/auth.ts
+++ b/src/backend/twitch-api/resource/auth.ts
@@ -27,7 +27,7 @@ export class TwitchAuthApi {
                     break;
             }
 
-            // This shouldnt only happen if an account is not logged in
+            // This should only happen if an account is not logged in
             if (!userId || !apiClient) {
                 return false;
             }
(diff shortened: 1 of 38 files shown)


src/backend/twitch-api/resource/auth.ts
```diff
@@ -27,7 +27,7 @@ export class TwitchAuthApi {
                     break;
             }
 
-            // This shouldn only happen if an account is not logged in
+            // This shouldnt only happen if an account is not logged in
```
Wissididom, line 30: Not sure, but to me, it seems like this should be either `shouldn't`, `should not` or `should`. If I understand the code correctly, I think it should be `should`.

CKY-: do you see any others?

CKY-: i stopped doing comments and started going after what is in the UI so there are lots more that are just comments.

zunderscore (approved):

PR #2326, opened 2024-01-04
fix OBS Source Screenshot regression

dennisrijsdijk:
### Description of the Change

This PR fixes a regression introduced into the OBS Source Screenshot effect in commit [3f5d04f](https://github.com/crowbartools/Firebot/commit/3f5d04fe871d7fa617e6cf18a49e2c6311873076) where effects created prior to this commit would no longer create screenshots.
It also adds validation to the effect model, ensuring at least one output is selected and applicable outputs aren't selected, but unconfigured.

### Testing

- Ensured an effect created before 3f5d04f saves the image to file again.
- Ensured effect model errors when no outputs are selected
- Ensured effect model errors when Folder, File or Discord are selected, but not filled in properly


ebiggz (approved):
Thanks for this. I added backwards compat to the frontend side but forgot the backend side clearly.

Issue #2187, opened 2023-06-20
[Feature Request] Pronoun variable 


CKY-:
can we have a variable that will return the viewers pronouns from pronoun picker, 
 if a pronoun is set by the user return it OR return empty and cache it for that sesion, 
 this can be use full for text strings to say that person did a thing 
 
$prounouns[target] =>[he/him]
$prounouns[target, 1] => he
$prounouns[target, 2] => him

because in grammar  you say "He did a thing" not "Him did a thing"

get the cached per session or ask the service to return a new.

JOELPOG:
so basically you want a variable where the users can set their pronouns like $setPronouns[he/him, she/her, they/them] and well this is stored in the DB and when the var is called you want to return the same pronouns which will have the format as 
```
$prounouns[target] =>[he/him]
$prounouns[target, 1] => he
$prounouns[target, 2] => him
```

the `target` placeholder being the username of the user...
is that right?

CKY-:
@JOELPOG 
> so basically you want a variable where the users can set their pronouns like $setPronouns[he/him, she/her, they/them] and well this is stored in the DB and when the var is called you want to return the same pronouns which will have the format as
> 
> ```
> $prounouns[target] =>[he/him]
> $prounouns[target, 1] => he
> $prounouns[target, 2] => him
> ```
> 
> the `target` placeholder being the username of the user... is that right?
Sorry I did not see this reply 
that is for the most part correct 

$prounouns[$target, subject] => he
$prounouns[$username, object] => him

but without the db part, firebot cashes the pronoun from a third party site that is an extension. 
to not get into Subject | Object using the array number would do the same thing. 

but when used in a command one can have an output that is 

go follow @JOELPOG $prounouns[$target, 1] is amazing. 

there are other uses of this as well 

this can be expanded into the realm of first second and third persons  

[Common personal pronouns in standard Modern English](https://mogai.miraheze.org/wiki/Pronoun)
Person | Number/Gender | Subject | Object | Dependent possessive (determiner) | Independent possessive | Reflexive
-- | -- | -- | -- | -- | -- | --
First | Singular | I | me | my |  mine | myself
 -- | Plural | we | us | our | ours | ourselves
Second | Singular | you | your | -- | yours | yourself 
 -- | Plural | --  | --  | -- | --  | yourself 
-- | -- | -- | -- | -- | -- | yourselves
-- | Masculine | he | him | his | -- | himself
[...]

PR #2341, opened 2024-01-06
#2187 V5 pronoun variable

Firebottle:
handles #2187

### Description of the Change
Creates a new replace variable that can pull pronoun information for a user from the pronoun site. It also allows the user to put in a fallback, as well as define which pronoun to use from a set like "They/Them".

$pronouns[$username, 0, they/them] = Returns "she/her" with fallback of "they/them"
$pronouns[$username, 1, they] = Returns "she" with fallback of "they"
$pronouns[$username, 2, them] = Returns "her" with fallback of "them"

### Applicable Issues
This won't cover every single possible combination of pronouns people might need of course. For example "his" is not something on the pronoun site, so people will need to get creative to cover some edge cases.

### Testing
- Create a command in firebot using the new variable.
- Run command in chat.
- Try putting gibberish in for the username to make sure fallbacks are used.


ebiggz (approved):
looks great!!

Issue #576, opened 2018-03-12
Chat Command Groupings for Quick Bulk-Activate/Disable State Changes


H0LD-MY-BEAR:
## What version of Firebot are you running?
Release v4.10.5

## Please describe the issue or feature requests in as much detail as possible.
Feature Request: Chat Command groupings that can be bulk toggled to Active/Disabled state - Example Concept: Subscriber Commands - If a streamer would like to disable a large selection of chat commands (their sub commands), the streamer currently needs to manually disable each command one at a time in a minimum of 3 clicks (open command -> uncheck 'active' -> save) or come up with a scripted way to recurse through a custom list of commands and activate/disable each one (I don't even know if that is supported, I'm just assuming).  Giving Firebot the ability to quickly disable an entire group of chat commands would give bot-users quick, easy control over an interactive feature they offer as a reward to special members of their community, who still, at times, abuse the reward at undesired times =)

ebiggz:
I can definitely see the need for this. 

I think the best way to accomplish this is to create an “Update Command” effect that, like the existing “Update Button” effect, would let you edit the active state of a given command. Then you could set up a hotkey or an interactive button to disable whatever commands you wanted.

Anabella00geek:
Addition of Groups so that commands, channel rewards, present effect lists, and timers could all be added to groups that could be turned off and on when needed. You could set up the group and add all the things that you would want to turn off with the click of a button within that one group.

C3l4h1r:
At it's core I was looking for something like this purely for an aesthetical, overview reason. Freely creatable folders to group commands or any trigger for that matter. Would really help and clean up some of the clutter I have right now with a growing list of triggers and events. Disabling triggers and the like on a folder based settings is a logical addition to this.

Oshimia:
I've had multiple people ask me for folders to sort their commands, and it's been raised again in the discord. This would be a very nice feature to have.

PR #2342, opened 2024-01-06
Toggle Commands, Channel Rewards, Time Based By Tags

dennisrijsdijk:
### Description of the Change

Adds the ability to use command tags to toggle:
- Commands
- Channel Rewards
- Timers
- Scheduled Effect Lists

### Applicable Issues

#576 

### Testing

Ensured for each effect that effect validation works properly.
Old effects still run as they did before these changes
New effects that toggle a single command, reward or time based thing still work
Toggle by tags works properly for each effect.
Requesting further testing by @CKY- and others

### Screenshots

image
image
image
image


ebiggz:
Quite a while ago now, Cave and I had a discussion about whether or not to use Sort Tags for this sort of thing. The consensus we came to at the time was that Sort Tags would not be used for any logic or functionality in the bot outside of sorting capabilities in the UI (hence their name). Instead, we would bring "Sets" (like what Events have) to every trigger type and Sets would have this sort of logic and functionality associated to them to stay consistent with Events.

While I still believe Sets would be more appropriate for this use case, I am willing to go back on the decision to not use Sort Tags for non-sorting functionality given where we are at in the life of v5 so we can at least get something close to the intended functionality to the end users, if everyone else is.

dennisrijsdijk:
temporarily marking as draft, awaiting possible conflicts with #2351

ebiggz (approved):
Good stuff here - thanks for taking it on!

Issue #2229, opened 2023-09-09
[Feature Request] Expand ExtraLife integration


CKY-:
**Describe the solution you'd like**
Iterate on the extralife integration to include support for:
- getting goal information
- getting milestones info
- top donator

API docs: https://github.com/DonorDrive/PublicAPI

ebiggz:
Updated title and description to include more details and link to api docs

Firebottle:
- Top X donors (with a warning that this often shows real names)
- X number of incentives ($10 - I play with one hand!)
- Incentive by dollar amount.
- Next milestone (get current raised amount and find the next milestone)
- Fundraising goal

PR #2343, opened 2024-01-07
feat(integrations): extra life variables

Firebottle:
Handles #2229 

### Description of the Change
This adds several new variables specifically for extra life. All of these have a "easy mode" where it'll just output a string with the information for easy display. However, you can also pass an argument to just get the JSON back for advanced usage.

extraLifeInfo = allows a user to basically get any extra life profile info. This includes things like goal amounts, avatarURL, etc.
extraLifeMilestones = allows the user to get milestone information.
extraLifeDonations = allows the user to get donation information.
extraLifeIncentives = allows the user to get incentive information.

### Applicable Issues
- All of these work, but I believe milestones and incentives could be made much better than what I have here.
- The node module has a "where" filter in it's docs, but I could never get a call to go through using it. If we could figure that out the code could be simplified quite a bit.

### Testing
Chat commands with each of these variables.


ebiggz (changes requested):
Awesome work on this!!

I would suggest an organizational tweak: move these variables into the `/integrations/builtin/extralife` area and have the extralife integration register them on init (similar to how the other integrations do with their effects & variables)

Additionally, as a quality-of-life improvement, we can take advantage of the participant ID that the users already provide in Settings > Integrations > ExtraLife so they don't have to provide it for every $variable. I think being able to provide a custom ID is cool, so perhaps we make the participant ID the last argument of each variable, making it easier to have it be optional and default to configured participant ID. 

You can get the configured participant ID like so:
```ts
if (!participantId) {
    participantId = integrationManager.getIntegrationAccountId("extralife");
}
```

ebiggz (approved):
Looks fantastic!

Issue #1947, opened 2022-09-22
[Bug] Command cooldowns resetting whenever command cache is refreshed


PonchoTaco:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am experiencing an issue where I have a command that has a user cooldown set for 28800s (8 hours) - basically I only want people to use it once per stream.

If I remain on the "Dashboard" tab, it works. The moment I go to the "Commands" tab and back to "Dashboard" tab, the cooldown resets and my chatters are able to use the command again.

### Expected Behavior

Cooldown should not reset when going between tabs in Firebot

### Steps To Reproduce

!command
Go to "Commands" tab
Go back to "Dashboard" tab
!command

### Version

5.54.2

### What operating system are you using Firebot on?

Windows 10

### Contact Details (Optional)

PonchoTaco#0001

PonchoTaco:
Actually, I'm able to replicate this doing multiple things - editing a different command, viewing other commands, different tabs, clicking all of these resets all my cooldowns

CKY-:
can confirm this

PonchoTaco:
Unfortunately, I need to move any commands (with cooldowns) I have over to a different bot as cooldowns are not useable on Firebot in the current state.

zunderscore:
This also occurs after the Toggle Command effect

MattyCanny:
This is actually quite frustrating, and it has taken me a long time to work out this is why my cooldowns are not working. 

As it has not been fixed though, I expect it is not as easy to fix as we would like.

zunderscore:
Yes, because of the way we currently load/cache commands, this would be a fairly complicated fix. We'll continue to investigate this, but it may not be until v6 when we can implement a solution.

PR #2351, opened 2024-01-08
chore(commands): refactor command manager, command handler

zunderscore:
### Description of the Change
Refactored the `CommandManager`  and `command-hander` (now `CommandHandler`) to TS, folded the `command-access` functionality into `CommandManager`, moved command data file access from the frontend exclusively to the backend, and added UI options for clearing cooldowns on commands, both individually and globally. Cooldowns no longer clear on command cache refresh, but only when they expire or when the user explicitly clears them.

### Applicable Issues
#1947

### Testing
- Validated creating, updating, and deleting custom commands from frontend
- Validated editing and resetting system commands from the frontend
- Validated all `!command` subcommands still work and reflect in the frontend
- Validated **Toggle Command** effect works as expected
- Validated cooldowns work and cache as expected
- Validated cooldowns do NOT clear when (re)loading the commands page
- Validated individual command cooldowns DO clear when selecting the context menu option
- Validated ALL command cooldowns clear when using the "Clear All Cooldowns" button

### Screenshots
image


src/backend/chat/commands/command-handler.ts
```diff
@@ -0,0 +1,640 @@
...
+        frontendCommunicator.send("command-count-update", {
+            commandId: command.id,
+            count: command.count
+        });
+    }
+
+    private parseCommandTriggerAndArgs(trigger: string, rawMessage: string, scanWholeMessage = false, treatQuotedTextAsSingleArg = false): TriggerWithArgs {
+        let args = [], rawArgs = [];
+
+        if (rawMessage != null) {
```
ebiggz, line 296: Super minor (and probably preexisting but just trying to be thorough): since`rawArgs` isn't needed in the scope its defined in, it can be moved within the `if(rawMessage != null) { }` block.

src/backend/chat/commands/command-handler.ts
```diff
@@ -0,0 +1,640 @@
...
+            const commandOptions = {};
+            if (command.options != null) {
+                for (const optionName of Object.keys(command.options)) {
+                    const option = command.options[optionName];
+                    if (option) {
+                        let value = option.value;
+                        if (value == null) {
+                            value = option.default;
+                        }
+                        commandOptions[optionName] = value;
```
ebiggz, line 403: I know this is preexisting stuff but for cleanups sake while we are in here, this could be modernized/simplified to:
```ts
commandOptions[optionName] = option.value ?? option.default
```

src/backend/chat/commands/command-manager.ts
```diff
@@ -0,0 +1,522 @@
...
+            this.deleteCustomCommand(command.id);
+        }
+    }
+
+    /**
+     * Refreshes the command cache
+     *
+     * @param retry The retry count. Default is `1`.
+     */
+    refreshCommandCache(retry = 1): void {
```
ebiggz, line 418: I'm pretty sure we can strip out the weird retry logic from this function now that we are read/writing data in the backend only. This was a hacky-ish solution to deal with weirdness steaming from the renderer and main services both trying to access the same file and we don't do it anywhere else so we shouldn't need it here.

ebiggz (changes requested):
Thanks for taking on a not-so-fun refactor / TS port, looks great!

I left a few comments in the spirit of cleaning things up while we are in here -  I realize most if not all are referencing old/preexisting code.

src/backend/chat/commands/command-handler.ts
```diff
@@ -0,0 +1,640 @@
...
+    private _handledMessageIds: string[] = [];
+
+    private buildCooldownCacheKey(commandId: string, subCommandId: string = null, username: string = null): string {
+        const tokens = [commandId];
+
+        if (subCommandId != null && subCommandId !== "") {
+            tokens.push(subCommandId);
+        }
+
+        if (username != null && username !== "") {
```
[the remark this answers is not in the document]
ebiggz: (this one may be controversial. I am a big fan of the double bang for checking for nullable/empty strings but also not a hill id die on)

ebiggz (approved):

Issue #1176, opened 2021-05-12
[Feature Request] RunProgram effect, Read output


CrowbarToolsRobot:
cky via Discord:

Having the ability to read in what applications are doing would be a nice addition. Sample: run console command and store result to variables  for later use.

zunderscore:
Released in v5.61.0

PR #2360, opened 2024-01-14
#1176 Add Read output from RunProgram Effect

alfw:
### Description of the Change
Added the stdout to a output `programResponse` so other effects can use the response from the program 

### Applicable Issues
- #1176

### Testing
- Calling own CLI tool for output and used a Custom Variable to set it
- Validated the text in Custom Variable Inspector was updated to expected output from own CLI tool.

### Screenshots
image
image


ebiggz (approved):
Nice work, thanks for taking this on!

PR #2397, opened 2024-02-09
feat: reuse error message when tokens are invalid

alfw:
### Description of the Change
After device flow auth changes there was no handling for when tokens are invalid / broken. The error message for the old system was still in code but not used anymore.

Added 2 functions to be able to set there is a issue with the tokens and use onRefreshFailure from DeviceAuthProvider that was not used.

### Applicable Issues
- None

### Testing
- Edit your auth-twitch.json to have a invalid access token and refresh token
- Start Firebot and message shows that your tokens are invalid 

### Screenshots
image


src/backend/common/account-access.js
```diff
@@ -215,16 +215,26 @@ frontendCommunicator.on("getAccounts", () => {
...
 });
 
+function setStreamerTokenIssue() {
+    streamerTokenIssue = true;
+}
+
+function setBotTokenIssue() {
+    botTokenIssue = true;
+}
+
```
zunderscore, line 230: To be consistent with other functions of this type, let's consolidate this down to a single function that takes an `accountType` parameter. Something like this should work:

```js
function setAccountTokenIssue(accountType) {
    if (accountType === 'streamer') {
        streamerTokenIssue = true;
    } else if (accounType === 'bot') {
        botTokenIssue = true;
    } else {
        throw new Error('invalid account type')
    }
}
```

src/backend/auth/firebot-device-auth-provider.ts
```diff
@@ -1,5 +1,5 @@
 import logger from "../logwrapper";
-import accountAccess, { FirebotAccount } from "../common/account-access";
+import accountAccess, { FirebotAccount, setBotTokenIssue, setStreamerTokenIssue } from "../common/account-access";
```
zunderscore, line 2: You shouldn't need to import any additional functions here. The default `accountAccess` object should contain what you need.

zunderscore (changes requested):

src/backend/auth/firebot-device-auth-provider.ts
```diff
@@ -82,6 +83,7 @@ class FirebotDeviceAuthProvider {
             });
 
             this.botProvider.onRefresh((userId, token) => this.onRefresh("bot", userId, token));
+            this.botProvider.onRefreshFailure(() => accountAccess.setAccountTokenIssue("streamer"));
```
zunderscore, line 86: This should pass `"bot"` as the parameter

zunderscore (changes requested):

zunderscore (approved):
Looks great, thanks!

Issue #2401, opened 2024-02-15
[Bug] quick actions are being indexed incorrectly 


CKY-:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when adding a quick action the entire quick action is removed 
because "position" is being duplicated. 

### Expected Behavior

"position" increment to the next value and it not break the ui 

### Steps To Reproduce

add a quick action. 

### Version

5.60.1

### What operating system are you using Firebot on?

Windows 11

CKY-:
fixed: 5.61.0

PR #2402, opened 2024-02-15
fix(quick actions): fix adding quick actions breaking

dennisrijsdijk:
### Description of the Change

Fix a regression with adding quick actions

### Testing

Ensured that adding quick actions through the UI and setups, and removing them through UI and Setups work properly.


CKY-:
#2401

zunderscore (approved):

PR #2403, opened 2024-02-15
chore: typos

CKY-:
### Applicable Issues

added missing letters from words 

mostly logging and chat messages 
1 UI change had a missing H

### DIFF
--- a/src/backend/chat/commands/builtin/quotes.ts
+++ b/src/backend/chat/commands/builtin/quotes.ts
@@ -305,7 +305,7 @@ export const QuotesManagementSystemCommand: SystemCommand<{
                         return resolve();
                     }
 
-                    await twitchChat.sendChatMessage(`Sorry! couldn't find a quote with that id number.`);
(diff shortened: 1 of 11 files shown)


src/backend/chat/commands/builtin/quotes.ts
```diff
@@ -305,7 +305,7 @@ export const QuotesManagementSystemCommand: SystemCommand<{
                         return resolve();
                     }
 
-                    await twitchChat.sendChatMessage(`Sorry! Couldnt find a quote with that id number.`);
+                    await twitchChat.sendChatMessage(`Sorry! couldn't find a quote with that id number.`);
```
zunderscore, line 308: `couldn't` should be `Couldn't` here

zunderscore (changes requested):

src/backend/chat/commands/builtin/quotes.ts
CKY-: updated Should that also be We couldn't since all others start that way?

zunderscore (approved):

Issue #2415, opened 2024-02-19
[Feature Request] User chat color variable


CrowbarToolsRobot:
cky_ via Discord:

Can we please have a $usernameColor[$username]
variable to output twitches user color code to be used in text and html effects?

zunderscore:
Released in v5.62.0

PR #2430, opened 2024-02-27
feat: $userChatColor

CKY-:
### Description of the Change

added user chat color variable
added associated Chat Api Endpoint 

### Applicable Issues

#2415 

### Testing

used in a chat message, preset  effect and event to verify that the user hex color was returned
returns White instead of string empty to not break CSS styling or other use cases.


src/backend/twitch-api/resource/chat.ts
```diff
@@ -328,4 +328,19 @@ export class TwitchChatApi {
...
         return false;
     }
+
+    /**
+     * Gets the chat color for a user.
+     *
+     * @param targetUserId numerical ID of the user as sting
+     * @returns the color as hex code, null if the user did not set a color, or undefined if the user is unknown.
+     */
+    async getColorForUser(targetUserId: string): Promise<string> {
```
dennisrijsdijk, line 338: the typescript return type should be Promise<string | null | undefined>

dennisrijsdijk (changes requested):

src/backend/variables/builtin/twitch/chat/message/chat-user-color.ts
```diff
@@ -0,0 +1,59 @@
...
+        possibleDataOutput: [OutputDataType.TEXT]
+    },
+    evaluator: async (trigger, username: string) => {
+        try {
+            let chatColor = "";
+            if (username !== null) {
+                const viewer = await viewerDatabase.getViewerByUsername(username);
+                if (viewer != null) {
+                    chatColor = await twitchApi.chat.getColorForUser(viewer._id);
+                    if (chatColor == null || chatColor === undefined) {
```
dennisrijsdijk, line 33: == null will also catch === undefined here.
This should be changed either to just == null or === null || === undefined

src/backend/variables/builtin/twitch/chat/message/chat-user-color.ts
```diff
@@ -0,0 +1,59 @@
...
+            }
+            if (trigger.metadata.chatMessage) {
+                chatColor = trigger.metadata.chatMessage.color;
+            } else if (trigger.type === EffectTrigger.EVENT || trigger.type === EffectTrigger.MANUAL) {
+                chatColor = trigger.metadata.eventData.color;
+            }
+            if (chatColor === "") {
+                const streamer = accountAccess.getAccounts().streamer;
+                chatColor = await twitchApi.chat.getColorForUser(streamer.userId);
+                if (chatColor == null || chatColor === undefined) {
```
dennisrijsdijk, line 47: == null will also catch === undefined here.
This should be changed either to just == null or === null || === undefined

dennisrijsdijk (changes requested):

Issue #2145, opened 2023-04-24
[Feature Request] Global toggle on /me for Firebot Messages


CrowbarToolsRobot:
Dennis.sh via Discord:

Add a global toggle for sending all chat messages (either from streamer, or bot account, or both) with /me for styling and accessibility purposes.

PR #2440, opened 2024-03-03
Me chat effect

heyaapl:
### Description of the Change

Added a new option to the Chat effect to apply `/me` to the Chat message or Whisper.

### Applicable Issues

[#2145](https://github.com/crowbartools/Firebot/issues/2145)

### Testing

Validated in both Chat messages and Whispers.

1. Add the Chat effect to a trigger (Command, Event, Effect List, etc)
2. Add whatever text you would like "This is a chat message" with or without a variable "This is a message today $date"
3. Check the "Use '/me'" checkbox
4. Trigger the effect and review your chat or whisper

### Screenshots

Screenshot 2024-03-03 at 10 52 50 AM


src/backend/effects/builtin/chat.js
```diff
@@ -19,6 +19,10 @@ const effect = {
         <textarea ng-model="effect.message" class="form-control" name="text" placeholder="Enter message" rows="4" cols="40" replace-variables></textarea>
         <div style="color: #fb7373;" ng-if="effect.message && effect.message.length > 500">Chat messages cannot be longer than 500 characters. This message will get automatically chunked into multiple messages if it's too long after all replace variables have been populated.</div>
         <div style="display: flex; flex-direction: row; width: 100%; height: 36px; margin: 10px 0 10px; align-items: center;">
+            <label class="control-fb control--checkbox" style="margin: 0px 15px 0px 0px"> Use "/me"<tooltip text="'Applies Italics to your Chat Message or your Chat Color if used in a Whisper'"></tooltip>
+                <input type="checkbox" ng-model="effect.me">
+                <div class="control__indicator"></div>
+            </label>
```
ebiggz, line 25: I know you are just mirroring the other checkboxes here, but worth mentioning that we now have a proper firebot checkbox component (created more recently) that we've been slowly refactoring to. Wouldn't hurt to use that here so we aren't adding additional instances of the "old way". Here's how to use it:
```html
<firebot-checkbox 
    label="Use \"/me\"" 
    tooltip="Applies Italics to your Chat Message or your Chat Color if used in a Whisper" 
    model="effect.me"
    style="margin: 0px 15px 0px 0px"
 />
```

src/backend/effects/builtin/chat.js
```diff
@@ -55,7 +59,10 @@ const effect = {
         } else if (trigger.type === EffectTrigger.EVENT) {
             messageId = trigger.metadata.eventData?.chatMessage?.id;
         }
-
+        const meCommand = "/me ";
+        if (effect.me) {
+            effect.message = meCommand.concat(effect.message);
+        }
```
ebiggz, line 65: We could use string template syntax here to avoid having to define a variable, like so:
```js
if (effect.me) {
    effect.message = `/me ${effect.message}`;
}
```

ebiggz (changes requested):
Real good stuff! Grats on the first PR. Detailed a couple small suggestions

src/backend/effects/builtin/chat.js
heyaapl: Perfect, thank you! I've gone through and updated the other checkboxes and the input box to use these Firebot Components in my latest commit.

src/backend/effects/builtin/chat.js
heyaapl: Thank you! I simplified the if logic and used the string template syntax

src/backend/effects/builtin/chat.js
```diff
@@ -16,26 +16,42 @@ const effect = {
...
+                model="effect.whisper"
+                style="margin: 0px 15px 0px 0px"
+            />
+            <div ng-show="effect.whisper">
+                <firebot-input 
+                    input-title="To"
+                    model="effect.whisperTarget" 
+                    placeholder-text="Username"
+                />
             </div>
```
ebiggz, line 45: Thanks for refactoring old html into the newer components! This change here however would constitute a breaking change for all chat effects created before this that had a whisper username set. This is because "whisperTarget" is a new property that didnt exist before, meaning no existing chat effect will have it set. 

Here's some tweaks we can make to prevent it from being a breaking change. 

1. For the `optionsController` in this file, we can emulate the same logic that was within the old "ng-init" property
```js
optionsController: ($scope) => {
    $scope.showWhisperInput = $scope.effect.whisper != null && $scope.effect.whisper !== ''
},
```

2. Then update the Whipser checkbox to point the model property to this new scope value we created in the controller:
```html
<firebot-checkbox 
    label="Whisper"
    model="showWhisperInput"
    style="margin: 0px 15px 0px 0px"
    ng-click="effect.whisper = ''"
/>
```

3. Update the firebot-input for the whisper username to point back to the original property:
```html
<firebot-input 
    input-title="To"
    model="effect.whisper"
    placeholder-text="Username"
/>
```

src/backend/effects/builtin/chat.js
```diff
@@ -56,7 +72,15 @@ const effect = {
...
-        await twitchChat.sendChatMessage(effect.message, effect.whisper, effect.chatter, !effect.whisper && effect.sendAsReply ? messageId : undefined);
+        if (effect.me) {
+            effect.message = `/me ${effect.message}`;
+        }
+
+        if (!effect.whisper && effect.whisperTarget != null) {
+            effect.whisperTarget = null;
+        }
+
+        await twitchChat.sendChatMessage(effect.message, effect.whisperTarget, effect.chatter, !effect.whisper && effect.sendAsReply ? messageId : undefined);
```
ebiggz, line 83: with the above changes made, these lines can be reverted back to what it was before, which was just:
```js
await twitchChat.sendChatMessage(effect.message, effect.whisper, effect.chatter, !effect.whisper && effect.sendAsReply ? messageId : undefined);
```

ebiggz (changes requested):

src/backend/effects/builtin/chat.js
heyaapl: Thank you for this! I've gone ahead and made these changes and pushed them up 🙏

src/backend/effects/builtin/chat.js
heyaapl: I've gone ahead and reverted this back to the original. Thank you again for your help!

ebiggz (approved):
Thanks for working on this Chuck!

PR #2447, opened 2024-03-06
fix:userChatColor string issue.

CKY-:
### Description of the Change

fixes string errors that where cropping up 
manages user retrieval better with regards to metadata received. 
adds all the null checking..  

### Testing

tested live with all the event types that could fail.


CKY-:
side note sorry i did not catch all this on the first pass..

src/backend/variables/builtin/twitch/chat/message/chat-user-color.ts
```diff
@@ -26,22 +26,22 @@ const model : ReplaceVariable = {
...
-                chatColor = trigger.metadata.eventData.color;
+                chatColor = trigger?.metadata?.eventData?.chatMessage?.color;
             }
-            if (chatColor === "") {
-                const streamer = accountAccess.getAccounts().streamer;
-                chatColor = await twitchApi.chat.getColorForUser(streamer.userId);
+            if (chatColor == null) {
+                const userId = trigger?.metadata?.userId ?? accountAccess.getAccounts().streamer.userId;
+                chatColor = await twitchApi.chat.getColorForUser(userId);
                 return chatColor ?? DEFAULT_COLOR;
```
dennisrijsdijk, line 45: This return seems redundant, merge with L47 and add the default color check there?

ebiggz (approved):
Great work sir!

Issue #2453, opened 2024-03-08
[Feature Request] Expand follow check restriction with follow duration


dennisrijsdijk:
**Describe the solution you'd like**
I'd like to be able to enter a minimum amount of time the user must be followed to all specified channels. 

**Additional context**
This would help in scenarios like giveaways where you may want to require a user be followed for a certain amount of time.

zunderscore:
Released in v5.62.0

Issue #1193, opened 2021-05-19
[Feature Request] Accept $streamer for the Follow Check restriction


CrowbarToolsRobot:
Perry via Discord:

Accept $streamer for the Follow Check restriction

ebiggz:
Just to be clear, this simple sounding request would require a massive overhaul of the variable system and likely wont happen for v5

brumoen:
Would it be easier/possible to just add a checkbox to check if the user is following the streamer account?
Or make it so that if field is left blank it defaults to check if user is following the streamer?

dennisrijsdijk:
> Would it be easier/possible to just add a checkbox to check if the user is following the streamer account? Or make it so that if field is left blank it defaults to check if user is following the streamer?

I feel like a checkbox would be better UX than leaving the box empty.

zunderscore:
Released in v5.62.0

PR #2456, opened 2024-03-08
Follow Restriction Improvements

dennisrijsdijk:
### Description of the Change
Adds the ability to check for the streamer's channel in the Follow restriction
Adds the ability to check for a certain follow age in the Follow restriction

### Applicable Issues

#1193
#2453

### Testing

Ensured the following:
- Updated UI for restriction works in commands and channel rewards
- A warning is now shown when custom follow check is used regarding the streamer/mod restriction from Twitch.
- By default, Custom follow check is selected.
- Commands and Channel Rewards will successfully reject their triggers when either the follow isn't met, or a follow restriction isn't met
- Both will fail gracefully (reject) if the Custom option is chosen, and a channel is entered that the streamer is not a moderator for
- Both will fail gracefully when an invalid username is entered into the Custom field
- Channel rewards will automatically reject if "Automatically approve/reject redemptions based on restrictions outcome" is enabled 
- The Follow Age setting only takes effect when enabled, and properly checks and handles the restriction.

### Screenshots

image
image

### DIFF
--- a/src/backend/common/user-access.js
+++ b/src/backend/common/user-access.js
@@ -13,26 +13,42 @@ const teamRolesManager = require("../roles/team-roles-manager");
 
 const followCache = new NodeCache({ stdTTL: 10, checkperiod: 10 });
 
-async function userFollowsChannels(username, channelNames) {
+async function userFollowsChannels(username, channelNames, durationInSeconds = 0) {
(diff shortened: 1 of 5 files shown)


ebiggz (approved):
Looks good! Well done

Issue #2504, opened 2024-04-04
[Feature Request] Oauth2 to have auth URL separate from token URL 


CKY-:
**Describe the solution you'd like**
Enter a clear and concise description of what you want to happen.
to follow [RFC6749](https://datatracker.ietf.org/doc/html/rfc6749) the Authorization server and Resource server while not always can/should be different endpoints.  while messing with oauth2, I ran into a situation that needed the auth server and the token server to be different..  firebot currently only accepts one endpoint for auth and token. 

**Additional context**

PR #2505, opened 2024-04-04
feat: Split auth and token endpoints if auth exist

CKY-:
### Description of the Change

there was a need by me to have the auth and token endpoints to be different in some use cases 
this simply adds the auth as an optional end point and if it exists then it use that instead of the token endpoint.
while i was editing the file there was a small linting issue that i resolved. 

### Applicable Issues

#2504 

### Testing

i have ran this for a full week reauthed with streamlabs and tippy stream and twitch(both main account and bot account) to see if any issues would arise.

### DIFF
--- a/src/backend/auth/auth-manager.ts
+++ b/src/backend/auth/auth-manager.ts
@@ -40,7 +40,7 @@ class AuthManager extends EventEmitter {
                 break;
 
             case "device":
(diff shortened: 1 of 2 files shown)


CKY-:
my fork is acting funny will resubmit.

Issue #2513, opened 2024-04-12
[Feature Request] $queueLength variable


dennisrijsdijk:
**Describe the solution you'd like**
a variable that returns the amount of items currently waiting in a queue (as shown in the UI on the queues page)

PR #2533, opened 2024-04-28
Feat: Add $effectQueueLength variable

codehdn:
### Description of the Change
I've added a new text variable "effectQueueLength" as a utility variable. This variable takes in a queue name 

### Applicable Issues

#2513

### Testing

- Setup a new effect queue call it "testing" make it whatever queue mode you like
- disable the effect queue
- Setup a command and call it !addtoqueue with an effect of Chat that says "hi"
- In the !addtoqueue command set the queue to be "testing" 
- Setup a command and call it !queuelength with an effect of Chat that says "$effectQueueLength[testing]"
- Setup one more command and call it !badqueue with an effect of Chat that says "$effectQueueLength[not-a-real-queue]"
- Now head to the dashboard 
- send a message of "!addtoqueue" which will bring the queue count to 1
- send a message of "!queuelength" it should return a queue length of 1
- send a message of "!addtoqueue" 3 times the new queue length should be 4 
- send a message of "!queuelength"  it should return a queue length of 4
- head over to the effect queues page and enable and re disable your queue
- head back to the dashboard
- send a message of "!queuelength" it should return a queue length less than 4 (exact number will depend on queue type and how long the queue was enabled, anything below 4 is a passing condition)
- send a message of "!badqueue" this queue does not exists and as such should return the text "Unknown"

### Screenshots

**Selecting the variable:**
Selecting the variable
**Primary use case:**
Primary use case
**Bad queue case:**
Bad queue case

### DIFF
--- a/src/backend/effects/queues/effect-queue-manager.js
+++ b/src/backend/effects/queues/effect-queue-manager.js
@@ -101,7 +101,7 @@ class EffectQueueManager extends JsonDbManager {
         }
 
         item.queue = effectQueueRunner.getQueue(itemId);
-
+        item.length = item.queue.length;
         return item;
     }
 
--- a/src/backend/variables/builtin/utility/get-effect-queue-length.ts
+++ b/src/backend/variables/builtin/utility/get-effect-queue-length.ts
@@ -1,0 +1,23 @@
+import { ReplaceVariable } from "../../../../types/variables";
+import { OutputDataType, VariableCategory } from "../../../../shared/variable-constants";
+
+import effectQueueManager from "../../../effects/queues/effect-queue-manager";
+
+const model : ReplaceVariable = {
+    definition: {
+        handle: "effectQueueLength",
+        usage: "effectQueueLength[text]",
+        description: "Returns the length of an effect queue. Useful for showing queue length in a command response.",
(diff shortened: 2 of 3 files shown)


src/backend/effects/queues/effect-queue-manager.js
```diff
@@ -101,7 +101,7 @@ class EffectQueueManager extends JsonDbManager {
         }
 
         item.queue = effectQueueRunner.getQueue(itemId);
-
+        item.length = item.queue.length;
```
codehdn, line 104: adding this so the express api also has easy access to this value. Its present when calling for all queues but not individual queues so assume just a small oversight.

dennisrijsdijk: When I implemented that originally, the paradigm was "fetch all queues, get length. Fetch specific queue, get the array of items instead for more advanced use"

the intention there was that if you needed the length, you'd get it from the array as the array is already present.

codehdn: Completely get it, I'm adding it for the less capable tools that can fetch against the api but only take the json key and don't permit any type of count/logic capabilities.

SReject (approved):

Issue #2519, opened 2024-04-20
[Bug] Api Request log message shows "server" IP rather than "client" IP


dennisrijsdijk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Firebot is running on machine A, API requests are made from machine B
The info log `info: API Request from: 192.168.10.230:7472, for path: /api/v1/status` shows IP A, rather than IP B

### Expected Behavior

the IP shown in the log line should be the IP of the machine making the API call (machine B), not the machine that hosts Firebot (machine A)

### Steps To Reproduce

1. Open Firebot on a computer
2. Find the IP of the computer on the network (eg. ipconfig)
3. Open the developer console in Firebot
4. On a different machine, go to `http://<ip>:7472/api/v1/status`
5. Look for the line "API request from..." and observe the IP is the Firebot computer's IP, not the client's IP

### Version

5.62.1

### What operating system are you using Firebot on?

Linux Distro, Windows 10

PR #2534, opened 2024-04-29
Fix: Log requestor ip for api calls

codehdn:
### Description of the Change

Changed from using req.headers.host to using req.socket.remoteAddress

### Applicable Issues

#2519 

### Testing

- Start firebot on this branch
- open the developer tools (you can also see this log in the terminal you started firebot from)
- make a request from another device to ${your_devices_local_ip}:7472/api/v1/status
- verify the logging is using the ip of the second device and not the device firebot is running on
- log should look like:  info: API Request from: 192.168.1.65, for path: /api/v1/status

### DIFF
--- a/src/server/api/v1/v1Router.js
+++ b/src/server/api/v1/v1Router.js
@@ -6,8 +6,13 @@ const logger = require("../../../backend/logwrapper");
 
 router.use(function log(req, res, next) {
     // here we could do stuff for every request if we wanted
+    const clientIp = req.socket.remoteAddress;
+    // check for proper ipv4 address (possible for ipv4 address in ipv6 format)
+    const matches = clientIp.match(/(\d+\.\d+\.\d+\.\d+)/);
+    // if we find a match it is ipv4 if not fall back to given client ip which is most likely ipv6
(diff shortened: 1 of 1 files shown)


src/server/api/v1/v1Router.js
```diff
@@ -6,8 +6,13 @@ const logger = require("../../../backend/logwrapper");
 
 router.use(function log(req, res, next) {
     // here we could do stuff for every request if we wanted
+    const clientIp = req.socket.remoteAddress;
+    // check for proper ipv4 address (possible for ipv4 address in ipv6 format)
+    const matches = clientIp.match(/(\d+\.\d+\.\d+\.\d+)/);
```
dennisrijsdijk, line 11: If anything, I'd want to know if a request went over ipv6 while being an ipv4 address, I've had dual-stack cause issues several times, and this change would make those more difficult to diagnose

codehdn: thats a solid point, I can remove the regex matching so that it just serves the raw remoteAddress. Through that enough information could be gathered to determine where the requests are coming from.

SReject (approved):

Issue #2537, opened 2024-04-29
[Feature Request] Update Inactive Viewer Time when using Channel Points


florian-luce:
**Describe the solution you'd like**
When a user use a Channel Points Reward, from the dashboard, the user always appears with the orange status (lurking)
While it is active and should appear green, depending on the time set in the options.

**Additional context**
1) User have orange status (is AFK - lurking)
2) User use Channel Point (no write in chat)
3) User is still have orange status (lurking) -> While it should be green here

codehdn:
I think I've got a fix for this but gave up my affiliate status last month and have no way to test the change. Following the code I do believe this should work, if anyone is able to take a look and verify, I will put up the pr once confirmed by someone who is affiliate. https://github.com/codehdn/Firebot/commit/1812e12bbadc1536417da0a3aa5822fe357930dc

PR #2542, opened 2024-05-02
Mark user active on channel point redemption

codehdn:
### Description of the Change
Marks a user active on channel point redemption. Updates RewardRedemptionMetadata type to include userId and userDisplayName, these then get checked on the triggerChannelReward method to make sure theyre truthy, if theyre not it is assumed the streamer triggered the use of the channel reward and as such doesnt mark the user active. 

### Applicable Issues

#2537 

### Testing

- I dont have a affiliate twitch account at this time so I have not personally tested this change

- For testing I would setup a channel reward 
- have a user join chat and wait for it to become inactive
- have that same user redeem a channel reward and verify they become active
- also need to test for when a user is not in the chat user list yet (CKY mentioned possible weirdness here)
- Need to test what happens when the redemption gets put into an effect queue, does it still mark user active or does it assume the streamer sent it in

Per CKY testing in the discord: 

" this change is working as advertised.
  there was a flutter when a user who was not in the list appeared after using a redeem 
 i would say PR it and let the core team have a look"
- https://discord.com/channels/372817064034959370/372822810168393728/1235656878990823517

### Screenshots

None


ebiggz (approved):
Looks good! Thanks for the contribution!

Issue #2589, opened 2024-06-28
[Bug] $readFile: Reading the last line of a text file does not work correctly


DontCallMeChicken:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

- Variable `$readFile[path\to\file.txt, last]` don't work.
Nothing is output when reading the last line of a text file.
- Variable `$readFile[path\to\file.txt, random]`.
Only outputs the lines before the last one. If the last line appears to be output, nothing is displayed. Just like with the _"last"_ option.

### Expected Behavior

The _"last"_ option should output the last line of the specified text file.
The _"random"_ option should also output the last line of the text file.

The variable `$readFile[path\to\file.txt]` outputs the entire text file completely and correctly.
These options read the text file correctly: _"first"_ and _"specific line number"_.

The variable `$fileLineCount[path/to/file.txt]` shows the correct number of existing lines in the text file.

### Version

5.62.1

### What operating system are you using Firebot on?

Linux Distro

### Relevant log output

```shell
No meaningful log entries available.
```

### Contact Details (Optional)

Discord: DCMC (dontcallmechicken.dcmc)

CKY-:
is the file ending in a new line ?

DontCallMeChicken:
> is the file ending in a new line ?

No, the end of the file, i.e. the last line, is also the last entry. Not an empty line.
I also tested it with other text files that have been around for a while and have always worked. The error also occurs there.

The effects and commands that use this variable no longer work either.

CKY-:
> * Variable `$readFile[path\to\file.txt, last]` don't work.
>   Nothing is output when reading the last line of a text file.
> * Variable `$readFile[path\to\file.txt, random]`.
>   Only outputs the lines before the last one. If the last line appears to be output, nothing is displayed. Just like with the _"last"_ option.

im going to ping you in discord i cant replicate this. lets see if we can diagnose it a bit?

DontCallMeChicken:
> im going to ping you in discord i cant replicate this. lets see if we can diagnose it a bit?

Okay. I'm excited! :)

phroggster:
Commit 895605a addressed this issue, and it's live in the nightly releases. Once the next release hits the proverbial store shelves with this change, the `$readFile` variable will have an additional option available to it for trimming out whitespace.

You'll be able to use it with:
- `$readFile[path\to\file.txt, first, true]` to get the **first** non-blank line
[...]

PR #2591, opened 2024-06-28
Address Issue 2589

phroggster:
### Description of the Change

Adds a `trimEnd()` to fileRead to help issues with reading an extra trailing blank line from a file.

### Applicable Issues

#2589

### Testing

Spent a good amount of time throwing files with random line feeds at this solution, and it seemed to be the least impact option available.

### DIFF
--- a/src/backend/variables/builtin/utility/file-read.ts
+++ b/src/backend/variables/builtin/utility/file-read.ts
@@ -53,7 +53,7 @@ const model : ReplaceVariable = {
             return contents;
         }
 
-        const lines = contents.split(/[\r\n]+/g);
+        const lines = contents.trimEnd().split(/[\r\n]+/g);
         if (Number.isFinite(Number(lineOrRandom))) {
             if (Number(lineOrRandom) <= lines.length) {
                 return lines[Number(lineOrRandom) - 1];


phroggster:
To expand a bit upon this one off to the side of the commit history, all of the available options for this are imprecise.

`trimEnd`'s drawback being that additional whitespace at the end of the last line of the file will also be removed. This may lead to issues where users expect that spacing to exist because "it's clearly there in the file", but it's suddenly not there but only for the last line upon read. This can be remedied by adding any desired spacing to wherever they're using the `evalVars` result, and removing it from their text file.

`trim` faced similar issues, with the difference being that leading whitespace was also removed from the file. It was not chosen, as most text editors do a very good job of showing leading newlines in a file, while also doing a poor job of showing (a) trailing newline(s).

Option C was to filter over the array of lines after the split to remove any blank ones, or rather, only allow non-empty elements. This one is probably the best option, but adds a significant execution cost to the variable expansion, depending on the length of the file. It's not too egregious, but it's enough to make me believe that `trimEnd` would be the best alternative. It's also needless, as only leading and trailing carriage returns/line feeds will trigger this bug. The split call will already handle removal of duplicate newlines in the middle of the file (but not if it's a line with only whitespace).

Option D is to solely check if the first and/or last element of the lines array is an empty string, and remove it from the array if so. This approach is three or six lines of code, and is reasonable. Certainly less costly than option C, none of the spacing issues that A and B entail, but it's still three or six lines of code.

Anyways, I hope this clears up why I selected this approach as the best option, as ten characters of source code and a single inexpensive call beats filtering an array of potentially massive size in my mind. Questions, comments, etc., please let me know, as I'm happy to explain myself further.

SReject:
Another option, and one that will cause less user impact, is that we could create an additional optional parameter to ignore empty lines and lines containing only whitespace.

```
$readFile[path, lineOrRandom, ignoreWhiteSpace]
```

phroggster:
Closing PR. SReject's got a backwards compatible plan that'll address this one in short order.

PR #2608, opened 2024-07-15
Feat: `obsIsConnected` Replace Variable

Oceanity:
### Description of the Change
Implementation of the replace variable `obsIsConnected` for use in conditionals to determine if Firebot should even bother attempting to use OBS-related Events and Websocket Requests without the need of the user to be Recording or Streaming.

### Applicable Issues
n/a

### Testing
Have manually tested this working in a local deploy of Firebot, it's literally just introducing a public getter and a variable interface to an already existing state in `obsRemote`, so not much room for weird edge cases where it could fail.

### Screenshots
n/a


src/backend/integrations/builtin/obs/variables/is-connected.ts
```diff
@@ -0,0 +1,15 @@
...
+export const IsConnectedVariable: ReplaceVariable = {
+    definition: {
+        handle: "obsIsConnected",
+        description:
+      "Returns 'true' if OBS is currently connected or 'false' if it is not.",
+        possibleDataOutput: ["text"]
+    },
+    evaluator: async () => {
+        const connectionState = isConnected();
+        return connectionState ?? false;
```
SReject, line 13: simplify this to

```js
return isConnected() ?? false;
```

src/backend/integrations/builtin/obs/variables/is-connected.ts
```diff
@@ -0,0 +1,15 @@
+import { ReplaceVariable } from "../../../../../types/variables";
+import { isConnected } from "../obs-remote";
+
+export const IsConnectedVariable: ReplaceVariable = {
+    definition: {
+        handle: "obsIsConnected",
+        description:
+      "Returns 'true' if OBS is currently connected or 'false' if it is not.",
```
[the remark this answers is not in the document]
Oceanity: Done! Not actually sure what did that, I don't remember manually adding it and I don't think it was prettier xP

src/backend/integrations/builtin/obs/variables/is-connected.ts
Oceanity: Done!

SReject (approved):

Issue #2611, opened 2024-07-16
[Feature Request] $userBio and or $userDescription variables


CrowbarToolsRobot:
cky_ via Discord:

Can we add $userBio and or $userDdescription variables twitch back it calls it description field of the user payload this will show allow users "bio" to be used in chat and overlays

PR #2612, opened 2024-07-16
Feat: `userBio` Replace Variable

Oceanity:
### Description of the Change
Implementation of the replace variable `userDescription` to pull the description (aka bio) of a user's profile. Went with `userDescription` over `userBio` to be consistent with internal terminology with Twitch.

### Applicable Issues
#2611 

### Testing
Manually tested use of `$userDescription` without a passed argument in all applicable Twitch Events and with the manual run button, and with both a manually provided and replace variable username in `$userDescription[username]`

### Screenshots
n/a


CKY-:
"Bio" is the front end terminology that twitch uses.

Oceanity:
> "Bio" is the front end terminology that twitch uses.

Only when editing it in Settings, "About" is how it's described on the UI, and "Description" internally.  I dunno, I feel like "Bio" is not the first term folks would intuitively think of to represent that data

SReject (approved):

PR #2618, opened 2024-07-17
Feat: `OBS Connected` and `OBS Disconnected` Events

Oceanity:
### Description of the Change
Little followup to the `obsIsConnected` Replace Variable, would like to be able to do Housekeeping on some sources/text when I first start up OBS and Disconnected is just good for keeping tabs on things

### Applicable Issues
n/a

### Testing
Deployed locally and both events fired as expected

### Screenshots
https://github.com/user-attachments/assets/0e22a576-29c1-4bd8-8671-da21a3bbf87f


ebiggz (approved):
Looks good, these will be useful events! Thanks for the continued contributions!

Issue #2399, opened 2024-02-14
[Feature Request] multiline text inputs  args for preset effects


CKY-:
**Describe the solution you'd like**
please update the ui input for run preset effects arguments to be multiline

PR #2626, opened 2024-07-19
feat: multiline textarea in preset list args (#2399)

dennisrijsdijk:
### Description of the Change

Updates the preset list argument inputs to textarea to allow multi-line input

These fields default to 1 line high, and can be changed by the user

### Applicable Issues

#2399 

### Testing

Ensured saved preset list args load into fields properly
Ensured args are saved properly into the effect
Ensured textareas start at 1 line when opening the effect

### Screenshots

image
image


ebiggz (approved):
great idea!

Issue #2576, opened 2024-06-10
[Bug] $randomActiveViewer[Role] fails to retrieve any user.


CrowbarToolsRobot:
dennisoti via Discord:

The role check fails as it still expects the viewer role to be a username array, rather than an array of id, username and displayName.

PR #2627, opened 2024-07-19
fix: $randomActiveViewer variable filtering

phroggster:
### Description of the Change

Corrects the behavior of the `$randomActiveViewer` variable to allow for filtering again, and adds better handling of `"null"` vs `$null` parameters.

The issue appears to have originated with 042cc567, when `CustomRolesManager.getRoleByName` migrated from returning an array of strings to an array of objects.

### Applicable Issues

#2576

### Testing
Pending, TBD.

### Screenshots
Not applicable.


ebiggz (approved):
Nice catch, thanks for the fix!

Issue #2617, opened 2024-07-17
[Feature Request] Aria is missing in some places


CKY-:
Please update aria tags to explain what element is being highlighted, 
all settings

PR #2642, opened 2024-07-28
feat: Update preset effect list aria item

CKY-:
### Description of the Change

adds title to the searchable combo
changes the name of Custom to Custom Effect List 
changed the name of Preset to Preset Effect List 
so that screen readers can know that Selected: Custom Effect List is selected. or Preset Effect List is selected 

### Applicable Issues

#2617

### Testing

i have used NVDA to verify what is being read out is what is displayed on the screen


ebiggz (approved):
Thanks for this CKY!

PR #2643, opened 2024-07-28
fix: Export Setup Components Filter Negation

phroggster:
### Description of the Change

- Fix the Export Setup Component dialog filtering. 
  - Angular will negate filters that start with an exclamation mark, which is cool. However, many commands start with an exclamation mark, so that negation behavior is undesired here.
  - This completely removes any leading exclamation marks from the filtering.
- Cleaned up a trivial arrow-parens lint warning.

### Applicable Issues

None applicable

### Testing

Tried a number of things to break it, but I think even this one will get past CK.

### Screenshots

Screenshot showing searching for (in clockwise order):
- Nothing
- Previously-negated text
- Group/Sorting tags
- Identifiers

command searching round 2


phroggster:
I love how I can rebase a PR to get it up to the HEAD of the base, but can't pull it back down locally. Sigh.

ebiggz (approved):

PR #2645, opened 2024-07-28
Fixed `possibleDataOutput` for `unixTimestamp` Replacement Variable

Oceanity:
### Description of the Change
Turns out Numbers are in fact Numbers, oops 😰

### Applicable Issues
n/a

### Testing
No functionality changes, but did confirm it still works just in case

### Screenshots
n/a


ebiggz (approved):
who woulda thunk a number is a number, but you're right

Issue #2635, opened 2024-07-23
[Feature Request] Effects for lerp transforming OBS Sources over time


Oceanity:
**Describe the solution you'd like**
Basically, allow a user to pick a Source and choose a start and end value to transform the scale/position/rotation lerped over a period of time with option to ease-in/ease-out

**Additional context**
This is a feature I used a bit in another bot, and through testing I can see that the way they accomplish it is sending a lerped value via OBS websocket request every frame (roughly 60 requests over a 1000ms transform)

PR #2664, opened 2024-08-04
Feature/obs animate transform

Oceanity:
### Description of the Change
This implements a new Effect, `Transform OBS Source`, that allows users to perform animated transforms to a source's Position, Scale or Rotation

### Applicable Issues
#2635 

### Testing
Have done fairly comprehensive testing of all fields and combination of fields, but would like another person to smoke test it in case I missed edge cases

### Screenshots
n/a


src/backend/integrations/builtin/obs/effects/transform-obs-source-scale.ts
```diff
@@ -0,0 +1,238 @@
...
+        const transformKeys = [];
+        if (effect.isTransformingPosition) {
+            transformKeys.push("positionX", "positionY");
+        }
+        if (effect.isTransformingScale) {
+            transformKeys.push("scaleX", "scaleY");
+        }
+        if (effect.isTransformingRotation) {
+            transformKeys.push("rotation");
+        }
```
ebiggz, line 217: I think increased type safety on the expected transform keys would be advantageous here for maintainability as it will help prevent whoopsies from other maintainers / our future selves.

Something like:
```ts
type TransformKey = "positionX" | "positionY" | "scaleX" | "scaleY" | "rotation";
```

And then use that anywhere where those keys are expected (eg `const parsedStart: Record<TransformKey, number> = {}`, `const transformKeys: TransformKey = [];`, etc)

(might want to define and export the type in `obs-remote.ts` since it needs to be utilized there too)

src/backend/integrations/builtin/obs/obs-remote.ts
```diff
@@ -785,6 +785,99 @@ export async function getAudioSources(): Promise<Array<OBSSource>> {
...
+        requestData: {
+            sceneName,
+            sceneItemId,
+            sceneItemTransform: transformStart
+        }
+    });
+
+    let time = 0;
+    let first = true; // Ensure end value is used even if duration is 0
+    while (first || time < duration) {
```
ebiggz, line 814: A `do...while` loop could be utilized here with the same effect, as the "do" block is executed before the first while check:
```ts
let time = 0;
do {
 //...
} while (time < duration);
```

ebiggz (changes requested):
I really appreciate how clean and easy to read your code is. I only have a couple of relatively minor suggestions for maintainability.

This is awesome though and I think users are going to have a lot of fun with it!

src/backend/integrations/builtin/obs/effects/transform-obs-source-scale.ts
```diff
@@ -0,0 +1,238 @@
...
+                input-type="number"
+                input-title="Duration"
+                placeholder-text="milliseconds"
+                model="effect.duration"
+                style="margin-bottom: 20px;" />
+            <div style="display: flex; gap: 20px;">
+                <label class="control-fb control--checkbox" style="flex-basis: 50%">Ease-In </tooltip>
+                    <input type="checkbox" ng-model="effect.easeIn" >
+                    <div class="control__indicator"></div>
+                </label>
```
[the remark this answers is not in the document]
Oceanity: Oooo, did not know that was a thing 👀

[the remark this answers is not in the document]
Oceanity: Done!

src/backend/integrations/builtin/obs/effects/transform-obs-source-scale.ts
```diff
@@ -0,0 +1,238 @@
...
+            );
+        };
+        $scope.getScenes();
+
+        $scope.getSources = (sceneName: string) => {
+            $scope.isObsConfigured = backendCommunicator.fireEventSync("obs-is-configured");
+
+            $q.when(
+                backendCommunicator.fireEventAsync("obs-get-transformable-scene-items", [sceneName])
+            ).then((sceneItems: OBSSceneItem[]) => {
```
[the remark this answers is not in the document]
Oceanity: Removed, love that's just baked in :D

src/backend/integrations/builtin/obs/obs-remote.ts
Oceanity: I always forget this syntax exists, it's so niche, but 100% perfect for this

src/backend/integrations/builtin/obs/effects/transform-obs-source-scale.ts
```diff
@@ -0,0 +1,238 @@
...
+        };
+
+        $scope.selectSceneItem = (sceneItem: OBSSceneItem) => {
+            $scope.effect.sceneItem = sceneItem;
+        };
+
+        $scope.getScenes = () => {
+            $scope.isObsConfigured = backendCommunicator.fireEventSync("obs-is-configured");
+
+            $q.when(backendCommunicator.fireEventAsync("obs-get-scene-list")).then(
```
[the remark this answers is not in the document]
Oceanity: Fixed!

src/backend/integrations/builtin/obs/obs-remote.ts
Oceanity: Done!

src/backend/integrations/builtin/obs/effects/transform-obs-source-scale.ts
Oceanity: Created a Key typing in `obs-remote`

Oceanity:
Added comment to techdebt the generic-ifying of the `getLerpedCallsArray` method, will contemplate some possible paths forward on that in the near future

Oceanity:
Also made some MINOR changes to the function itself, noticed a few unwanted behaviors, but it seems clean now 😎 

Now if `duration` is `0` or `undefined`, it will just instantly transform the source to the end transform, or start transform if end is not provided

ebiggz (approved):
Looking fantastic, thanks for all the time and effort you put into this!

Issue #2634, opened 2024-07-23
[Feature Request] activity feed effect 


CKY-:
**Describe the solution you'd like**
please and an activtyfeed effect to be used from anywhere to add user defined data to the activity feed 
**Additional context**

phroggster:
https://github.com/phroggster/firebot-notification-examples

PR #2667, opened 2024-08-05
feat (effect): add activity feed effect

CKY-:
### Description of the Change

this effect adds the ability to create entries in the activity feed.  

this change Required UUID Types so that was added in the package it is working but not sure if version 10.0.0 is too new? 
im unsure if the package lock is needed. 

### Applicable Issues

#2634

### Testing

added effect to preset list and changed the icon added text then clicked play 
added effect to preset list left icon empty and added text then clicked play as noted in the screenshots 

### Screenshots

image
image
image

### DIFF
--- a/src/backend/effects/builtin/activity-feed-alert.ts
+++ b/src/backend/effects/builtin/activity-feed-alert.ts
@@ -1,9 +1,6 @@
 import { EffectType } from "../../../types/effects";
 import { EffectCategory } from "../../../shared/effect-constants";
-import moment from "moment";
-import { v4 as uuid } from "uuid";
-import frontendCommunicator from "./../../common/frontend-communicator";
-
+import { handleTriggeredEvent } from "../../events/activity-feed-manager";
 
 const effect: EffectType<{
     message: string;
@@ -39,22 +36,24 @@ const effect: EffectType<{
         return errors;
     },
     onTriggerEvent: async ({ effect }) => {
-
-        frontendCommunicator.send('event-activity', {
-            message: effect.message,
-            icon: effect.icon || "fad fa-comment-exclamation",
-            acknowledged: false,
-            event: {
-                id: "activity-feed-alert",
-                name: "Activity Feed Alert"
-            },
-            id: uuid(),
-            source: {
+        handleTriggeredEvent(
+            {
                 id: "firebot",
                 name: "Firebot"
             },
-            timestamp: moment().format("H:mm")
-        });
+            {
+                id: "activity-feed-alert",
+                name: "Activity Feed Alert",
+                activityFeed: {
+                    icon: effect.icon || "fad fa-comment-exclamation",
+                    getMessage: () => {
+                        return effect.message;
+                    }
+                }
+            },
+            {
+                username: "firebot"
+            });
 
         return true;
     }
--- a/src/backend/effects/builtin/activity-feed-alert.ts
+++ b/src/backend/effects/builtin/activity-feed-alert.ts
@@ -42,7 +42,7 @@ const effect: EffectType<{
 
         frontendCommunicator.send('event-activity', {
             message: effect.message,
-            icon: effect.icon ?? "fad fa-tv-alt",
+            icon: effect.icon || "fad fa-comment-exclamation",
             acknowledged: false,
             event: {
                 id: "activity-feed-alert",
--- a/src/backend/effects/builtin/activity-feed-alert.ts
+++ b/src/backend/effects/builtin/activity-feed-alert.ts
@@ -16,15 +16,25 @@ const effect: EffectType<{
     },
     optionsTemplate: `
     <eos-container>
-        <p>Use this effect to send yourself alerts in Firebot's activity feed. This alert is only visible to you if the activity feed is visible.</p>
+        <p>Use this effect to send yourself alerts in Firebot's activity feed.</p>
     </eos-container>
-    <eos-container header="Alert Icon" pad-top="true">
-        <h3>Icon</h3>
-        <p class="muted">A custom icon which allows you to identify your Activity alert.</p>
-        <input maxlength="2" type="text" class="form-control" ng-model="effect.icon" icon-picker required>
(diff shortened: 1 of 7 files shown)


src/backend/effects/builtin/activity-feed-alert.ts
```diff
@@ -0,0 +1,64 @@
...
+        id: "firebot:activity-feed-alert",
+        name: "Activity Feed Alert",
+        description: "Display an alert in Firebot's activity feed",
+        icon: "fad fa-comment-exclamation",
+        categories: [EffectCategory.FUN],
+        dependencies: []
+    },
+    optionsTemplate: `
+    <eos-container>
+        <p>Use this effect to send yourself alerts in Firebot's activity feed. This alert is are only visible to you if the activity feed is visible.</p>
```
phroggster, line 23: This language... My eyes are bleeding!

src/backend/effects/builtin/activity-feed-alert.ts
```diff
@@ -0,0 +1,64 @@
...
+        if (effect.message == null || effect.message === "") {
+            errors.push("Alert message can't be blank.");
+        }
+        return errors;
+    },
+    onTriggerEvent: async ({ effect }) => {
+
+        frontendCommunicator.send('event-activity', {
+           message: effect.message,
+            icon: effect.icon ?? "fad fa-tv-alt",
```
phroggster, line 46: Empty string icon will bypass this

phroggster (commented):
Couple of minor things worth poking

src/backend/effects/builtin/activity-feed-alert.ts
CKY-: no it dont

CKY-: image

src/backend/effects/builtin/activity-feed-alert.ts
```diff
@@ -0,0 +1,62 @@
...
+                activityFeed: {
+                    icon: effect.icon || "fad fa-comment-exclamation",
+                    getMessage: () => {
+                        return effect.message;
+                    }
+                }
+            },
+            {
+                username: "firebot"
+            });
```
ebiggz, line 56: Given the changes we made to the `handleTriggeredEvent` function, we can pass those bools here as the fourth arg:
```js
    {
        forceAllow: true,
        canRetrigger: false
    });
```

src/gui/app/services/activity-feed.service.js
```diff
@@ -24,7 +24,7 @@
                 }
 
                 const allowedEvents = settingsService.getAllowedActivityEvents();
-                if (!allowedEvents.includes(`${activity.source.id}:${activity.event.id}`)) {
+                if (!allowedEvents.includes(`${activity.source.id}:${activity.event.id}`) && `${activity.source.id}:${activity.event.id}` !== "firebot:activity-feed-alert") {
                     return;
                 }
```
ebiggz, line 29: Given the changes we made to pass along event settings, we can remove the hardcoded check and do this instead:
```js
if (!activity.event.forceAllow && !allowedEvents.includes(`${activity.source.id}:${activity.event.id}`)) {
    return;
}
```

src/gui/app/templates/chat/_chat-messages.html
```diff
@@ -270,6 +270,7 @@ <h4><i class="fas" ng-class="currentThreadMessages().length < 2 ? 'fa-reply' : '
           style="font-size: 15px; display: flex; align-items: center"
         >
           <i
+            ng-if="activity.source.id + ':' + activity.event.id !== 'firebot:activity-feed-alert'"
```
ebiggz, line 273: Given the changes we made to pass along event settings, we can remove the hardcoded check and do this instead:
```html
ng-if="activity.event.canRetrigger !== false"
```
(doing an explicit false check here just to ensure the default is showing the retrigger button)

src/backend/effects/builtin/activity-feed-alert.ts
```diff
@@ -0,0 +1,62 @@
...
+        <p>Use this effect to send yourself alerts in Firebot's activity feed. This alert is only visible to you if the activity feed is visible.</p>
+    </eos-container>
+    <eos-container header="Alert Icon" pad-top="true">
+        <h3>Icon</h3>
+        <p class="muted">A custom icon which allows you to identify your Activity alert.</p>
+        <input maxlength="2" type="text" class="form-control" ng-model="effect.icon" icon-picker required>
+    </eos-container>
+    <eos-container header="Alert Message" pad-top="true">
+        <textarea ng-model="effect.message" class="form-control" name="text" placeholder="Enter message" rows="4" cols="40" replace-variables></textarea>
+    </eos-container>
```
ebiggz, line 28: Couple notes here:
- Since the icon is optional and the least important field here, I think it should be moved after the Message section to de-emphasize it. 
- I don't think there is a need for two different Icon headers, so we can remove the `<h3>` in favor of the built in one from `<eos-container>`.
- Instead of the `<textarea>`, we can use a `<firebot-input>` component

Here's what all those tweaks would look like if you want an easy copy/paste:
```html
<eos-container header="Message" pad-top="true">
        <firebot-input 
            model="effect.message" 
            placeholder-text="Enter message"
            use-text-area="true"
            rows="4"
            cols="40"
        />
</eos-container> 
<eos-container header="Icon" pad-top="true">
        <input 
			maxlength="2" 
			type="text" 
			class="form-control" 
			ng-model="effect.icon" 
			icon-picker required
		/>
</eos-container>
```

ebiggz (changes requested):
Thanks so much for working on this! 

Couple feedback items to tighten things up and set us up better in the future. Should be good to go after that! 

I know you are taking a break at the moment so if you'd prefer that I go ahead and apply these tweaks, just ping or DM me in Discord 👍🏻

ebiggz (approved):
This is great, thank you CKY!

Issue #2566, opened 2024-05-28
[Bug] ObsSceneItemName returns unknown


CrowbarToolsRobot:
cky_ via Discord:

When an item is in a group 
ERROR Failed to get OBS scene items in scene "Scene Overlay Group"
and unknown is returned

CKY-:
In obs add a group to sources add an image to the group 

in firebot add a 
OBS Scene Item Enable State Changed event 
log out 
$obsSceneItemName

change visibility on the source 
ERROR Failed to get OBS scene items in scene "Name of Group"

unknown will be returned for the name

PR #2674, opened 2024-08-09
OBS Groups, sigh

phroggster:
### Description of the Change

This is one potential implementation of dealing with OBS groups, which are like OBS scenes, but like, not even close.

There are two (or more) potential ways to "solve" this issue:
1. Cache OBS group and scene information.
  a. Uses more memory, but is less IPC/network intensive, and more responsive.
2. Recursively look up and parse group information when specific events are received.
  a. Uses less memory, but is more IPC/network intensive, and slows down event propagation.

This PR uses approach 1, 

**DO NOT MERGE THIS INTO THE PENDING RELEASE, FOR THE LOVE OF $DEITY**

### Applicable Issues

#2566 

### Testing

Not nearly enough. **Have at it, please.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "firebotv5",
-    "version": "5.63.0-beta2",
+    "version": "5.63.0-beta3",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "firebotv5",
-            "version": "5.63.0-beta2",
+            "version": "5.63.0-beta3",
             "license": "GPL-3.0",
             "dependencies": {
                 "@aws-sdk/client-polly": "^3.26.0",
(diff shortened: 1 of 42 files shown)


phroggster:
This isn't a major PR, as it only *potentially* breaks filters, and none of these items actually have filters written for them.

phroggster:
So I messed up and based the actual work here upon the wrong branch, skipping over the fact that I never addressed the actual issue in it: `$obsSceneItemName` isn't cool with being sent a group. Now it is. The log messages are triggered on an OBS Scene Item Enable State Changed event, with the message text of
> `OBS item enabled state changed in scene $obsSceneName; { group_item_id: $obsGroupItemId, group_name: $obsGroupName, enabled: $obsSceneItemEnabled, scene_item_id: $obsSceneItemId, scene_item_name: $obsSceneItemName }`

OBS scene layout:
obs groups

Log messages while toggling the highlighted "eyeballs":
firebot-logs

ebiggz (approved):
all hail phroggster for this pr

PR #2675, opened 2024-08-09
fix: OBS Variable Typing

phroggster:
### Description of the Change

    - $obsInputMonitorType: object -> text
    - $obsIsConnected: text -> bool
    - $obsIsRecording: text -> bool
    - $obsIsStreaming: text -> bool
    - $obsSceneItemEnabled: text -> bool
    - $obsSceneItemName: number -> text

### Applicable Issues

N/A


phroggster:
This isn't a major PR, as it only potentially breaks filters, and none of these items actually have filters written for them.

ebiggz (approved):

PR #2688, opened 2024-08-13
chore: update error messages and verbiage for Macro Variable Arguments

dennisrijsdijk:
### Description of the Change

Adds more granular error messages and updates the verbiage around macro variable arguments

### Testing

Ensured error messages show up only when their condition check fails
Ensured macro vars only save when all conditions pass

### Screenshots

image
image
image
image
image


ebiggz (approved):
Thanks for cleaning this up!

Issue #2594, opened 2024-07-03
[Feature Request] Add Additional Option for the Play Video effect: Featured Clips


CrowbarToolsRobot:
angsaysroar via Discord:

It would be great if we could have another options added in the Play Video effect to only show featured clips when doing shoutouts.

Twitch loves to pull older clips or clips that streamers may not be something the streamer would choose, so having the bot only pulled from a streamer's featured clips would be great!

Issue #1554, opened 2022-01-21
[Feature Request] Date ranger filter on Random Clips


CrowbarToolsRobot:
CaveMobster via Discord:

A date range filter would be nice for the Random Twitch Clips option in the Play Video Effect, because what if you get that super awkward clip from one of your first streams... For everyone to see.... On stream...

PR #2695, opened 2024-08-14
feat: twitch clip date and feature

CKY-:
### Description of the Change

this adds 2 check boxes to the play random clip from twitch 
one for featured only 

one for max age with max age the time input was added so that users can select the amount of time to the current time to get clips from the api request.

and fixed linting
### Applicable Issues

#2594
#1554

### Testing

while coding and playing with this i was randomly targeting chat users 

### Screenshots

image

### DIFF
--- a/src/backend/effects/builtin/play-video.js
+++ b/src/backend/effects/builtin/play-video.js
@@ -124,6 +124,38 @@ const playVideo = {
                 model="effect.twitchClipUsername"
                 placeholder-text="Ex: $streamer, $user, etc"
             />
+            <div
+                class="mt-6 mb-2"
+                ng-class="{'has-error': $ctrl.formFieldHasError('clipSeconds')}"
+            >
+                <div class="form-group flex-row jspacebetween" style="margin-bottom: 0;">
+                    <div>
+                        <label class="control-fb control--checkbox" style="margin:0;"> Maximum Clip Age
+                        <input type="checkbox" ng-model="effect.useMaxClipAge">
+                            <div class="control__indicator"></div>
+                        </label>
+                    </div>
+                </div>
+                <div class="mt-2">
+                    <time-input
+                        ng-model="effect.maxClipAge"
+                        min-time-unit="'Days'"
+                        name="clipSeconds"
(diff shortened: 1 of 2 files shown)


src/backend/effects/builtin/play-video.js
```diff
@@ -124,6 +124,38 @@ const playVideo = {
...
+                class="mt-6 mb-2"
+                ng-class="{'has-error': $ctrl.formFieldHasError('clipSeconds')}"
+            >
+                <div class="form-group flex-row jspacebetween" style="margin-bottom: 0;">
+                    <div>
+                        <label class="control-fb control--checkbox" style="margin:0;"> Maximum Clip Age
+                        <input type="checkbox" ng-model="effect.useMaxClipAge">
+                            <div class="control__indicator"></div>
+                        </label>
+                    </div>
```
ebiggz, line 137: Please use the `<firebot-checkbox />` component for this bad boy. I know theres probably other checkboxes in this effect not using the "old way", I'm just trying to make sure at least that any new checkboxes we add are using the component :)

src/backend/effects/builtin/play-video.js
```diff
@@ -124,6 +124,38 @@ const playVideo = {
...
+                </div>
+                    <div class="mt-6 form-group flex-row jspacebetween" style="margin-bottom: 0;">
+                        <div>
+                            <label class="control-fb control--checkbox" style="margin:0;"> Only Featured Clips
+                                <input type="checkbox" ng-model="effect.isFeatured">
+                                <div class="control__indicator"></div>
+                            </label>
+                        </div>
+                    </div>
+            </div>
```
ebiggz, line 158: Can use `<firebot-checkbox />` here as well

ebiggz (changes requested):

SReject (approved):

ebiggz (approved):

angsaysroar:
@CKY- this is look really spiffy!

angsaysroar:
I noticed it doesn’t account for when someone does not have featured clips, so maybe add else if clause if the effect has thee maximum age of a clip play that else play a random clip. 

I hope that made sense. Lol

PR #2728, opened 2024-08-22
chore: Twitch Poll Length Warnings

phroggster:
### Description of the Change

Add some "better" UI hints about Twitch Poll length limits.

- This is a hassle, as variables are allowed, but can't be size-limited.
- So, warn where variables are used (rather, any dollar sign).
- But error where no dollar signs are used and size limits are exceeded...
- ...In the entry UI, the validator, **and** the add/edit choice modal.

### Applicable Issues

Discord issue: [Poll effect not triggering](https://discord.com/channels/372817064034959370/1276017580737298443)

### Screenshots

In no particular order:

- Title length limit exceeded (no variable):
title flat limit

- Title warning limit (with variable):
title var limit

- Choice warning limit (with variable):
choice var limit

- Empty Edit choice dialog:
Empty edit choice

- Add choice error (no variable):
add choice flat limit

- Add choice non-unique (variable):
add must unique


ebiggz (approved):
Thanks for implementing this! Having to account for variables makes length validation a pain but I like how you dealt with it here

Issue #2727, opened 2024-08-22
[Bug] prediction events are not retning metadata


CKY-:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

metadata is being dropped in the prediction event 

### Expected Behavior

the winning outcome should be returned 

### Steps To Reproduce

add a prediction ended event to it use the winningpredictionoutcome in a log and the variable is empty 
i have tracked it to the event router this line is the last line that i can find the information in. and i got lost so did the metadata
/v5/src/backend/events/events-router.js#L123

### Version

9.63 beta3

### What operating system are you using Firebot on?

Windows 10

PR #2729, opened 2024-08-22
fix: effect runner discarding context 

CKY-:
### Description of the Change

add references to trigger event and output so the context remains

### Applicable Issues

#2727

### Testing

using predictions to get test the data is infact present on the effects


ebiggz:
@CKY- Thanks so much for looking into this! After digging into it some more as well, I came to the conclusion we should just start using a vetted deep cloning library which handles various js wonkiness for us, because this sort of thing would definitely bite us again. Especially with twurple using an unorthodox pattern of custom symbols with getters.
https://github.com/crowbartools/Firebot/commit/7a6da1e399f01bc805f192d19a3c79e8926a7763

Issue #2353, opened 2024-01-09
[Feature Request] Default Chat account setting for dashboard


CrowbarToolsRobot:
ballzzpta via Discord:

Add a setting to allow which account(streamer/bot) to be the default chatter on the dashboard. Currently, if you select the bot as the chatter, then navigate from and then back to the dashboard, the chatter will revert to the streamer.

phroggster:
My proposed solution is to remember the current chat sender while you're navigating around the app. It won't persist this after restarting the application, which is going to fall a smidge short of what's being asked for here, but feels to provide for a better UX. Meanwhile, adding an actual option for the default chatter feels like settings bloat.

Hopefully this will provide for a middle ground acceptable to all, while minimizing confusion and settings fatigue?

phroggster:
I one-upped my prior comment, and changed the PR to remember the chat sender, chat message, and reply-to information. You can now hit reply on a message, switch the account you're messaging from, type out of a lengthy response, then change tabs to fix that one trivial mistake in an event that popped up while you were typing, and go back and resume typing. It'll retain the chat sender, chat message, and reply-to for you automatically. It still will not retain any of this information after restarting the application for the reasons given above. Expect this functionality to be included in the next release when it comes out.

PR #2772, opened 2024-08-31
feat: Remember Active Dashboard Chat User, Message Text, and Replying-To

phroggster:
### Description of the Change

- Caches the active chat sender, outgoing chat message text, and reply/thread information in the dashboard when the chat-messages controller gets reloaded.
- This intentionally **does not** get saved to disk, and the application will always start back up with "Streamer" as the active chat sender.
- Does not add any actual settings, checkboxes, nor UI. It just remembers the thing you already changed while the application stays active.
- Previously, changing component tabs then changing back would cause the chat sender to get reset to Streamer, the message text to be blank, and the reply-to/thread information to be dropped. This was *not* cool.

### Applicable Issues

#2353

### Testing

Yup, video evidence of functionality:
[Video Of Tests](https://github.com/user-attachments/assets/befc6419-af88-44f3-a308-9f4d2a7ee129)

### Screenshots

N/A

### DIFF
--- a/src/gui/app/controllers/chat-messages.controller.js
+++ b/src/gui/app/controllers/chat-messages.controller.js
@@ -11,15 +11,11 @@
             utilityService,
             activityFeedService
         ) {
+            $scope.afs = activityFeedService;
+            $scope.cms = chatMessagesService;
             $scope.settings = settingsService;
 
-            $scope.afs = activityFeedService;
-
-            $scope.chatMessage = "";
-            $scope.chatSender = "Streamer";
             $scope.disabledMessage = "";
-
-            $scope.cms = chatMessagesService;
 
             $scope.selectedUserData = {};
 
@@ -126,7 +122,7 @@
             };
 
             $scope.updateChatInput = function(text) {
-                $scope.chatMessage = text;
+                chatMessagesService.chatMessage = text;
                 focusMessageInput();
             };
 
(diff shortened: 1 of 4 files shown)


Wrongtown:
Nice one, I noticed this in my re-testing of changes for #2753 👍

ebiggz (changes requested):
This is a much needed improvement, thanks for doing it!

One minor suggestion: since we just want to cache UI state in memory, I think it would make most sense to store the `chatSender` variable in the associated `chat-messages.service` instead of using the settings service. "Settings" in this context currently represents any user defined settings that we want to persist to file, so having a memory only setting here to store UI state feels outside the scope.  In the event we do want to persist the selected chatter to file though, we could move to using settings!

General implementation:
```js
// chat-messages.service.js
service.chatSender = "Streamer";
```

```js
// chat-messages.controller.js
$scope.setChatSender = function(chatter) {
    chatMessagesService.chatSender = chatter;
};
```

```html
<!-- _chat-messages.html (chat messages service already passed to scope as $scope.cms)  -->
<span class="ml-4" style="width: 100%; text-align: center">{{cms.chatSender}}</span>
```

phroggster:
Thanks for the nudge away from `settingsService`, that did seem like the wrong place for this. I'm going to take your wisdom into account, but also add caching for any pending message text. The breadth of this PR has now been bumped up a notch, ever since I typed out a wordy message then switched tabs to look up an identifier and lost my message text the other day. 😭

Should have this PR updated within a few hours.

ebiggz (approved):
Ooo great idea to update reply-to and message text as well. This will be so nice. Thanks for rockin' this!

Issue #2775, opened 2024-09-02
[Feature Request] `arraySlice` replace variable to quickly get a truncated version of an array


Oceanity:
**Describe the solution you'd like**
A replace variable that takes an array, a start index, and an end index and returns a truncated array of just those elements

**Additional context**
n/a

PR #2776, opened 2024-09-02
Feat: `arraySlice` replace variable

Oceanity:
### Description of the Change
Adds an `arraySlice` replace variable that lets users easily get a truncated version of an array, also fixed a typo

### Applicable Issues
#2775 

### Testing
Tested the variable with various starts/ends including positive integers, negative integers, decimals, invalid strings, empty strings and undefined, including not including either start or end, and just including start

### Screenshots
n/a


ebiggz (approved):
LGTM 👍🏻

Issue #2348, opened 2024-01-08
[Feature Request] Question & All Answer variables for polls & predictions


CrowbarToolsRobot:
ollyrobzttv via Discord:

Could I ask that for the new poll (and prediction) based event triggers, we can also get the poll question as a variable, and all answer choices and results as an array? The main use case I see for "Poll Started" as an event is being able to show at least the question within the stream overlay somewhere, and personally my goal is to show the answers and results in real time until the poll ends - at the moment i can only show the current most popular answer and no way to pull in the question. Thank you as always! 😄

phroggster:
In the next release of Firebot, `$pollChoices` will have been added such that you can access the poll choice information, and `$pollTitle` will give you the (*wait for it*) poll title inside of the Poll Started, Poll Progress, and Poll Ended events. I need to revue to see if we're ready to do something similar for predictions, or if we're still held back by some missing user-interface code needed by that.

- `$pollTitle` is straightforward, and will simply return the string of text representing the title of the poll.
- `$pollChoices` is a bit more complex, as it is an array of objects representing the poll choices, but it provides for fairly simple access to these objects, and includes the examples pictured below. Such that:
  - `$pollChoices`, or `$pollChoices[raw]` will return an array that resembles:
```js
[
  { title: "Option A", index: 1, points: 690, votes: 69 },
  { title: "Choice B", index: 2, points: 750, votes: 75 },
  { title: "Polls need three options!", index: 3, points: 10, votes: 1}
]
```
  - `$pollChoices[count]` will give you a number representing the quantity of available poll choices. This will yield `3` from the example above.
  - `$pollChoices[title]` will give you the array of poll choice title strings, in the order that they were presented to users. This will yield `[ "Option A", "Choice B", "Polls need three options!" ]` from the example above.
  - `$pollChoices[votes]` will give you an array of numbers representing the number of votes cast for each choice, in the order that they were presented to users. Yields: `[69, 75, 1]`
  - `$pollChoices[points]` will give you an array of the number of channel point votes cast for each choice, in the order that they were presented to users. Yields: `[690, 750, 10]`
  - You can add a second parameter of either "index' or "votes" to sort the array(s) differently. The default sort is `index`, (or `null` / `$null`) and returns the items in the same way that they were presented to users. The `votes` sort method will arrange from most-voted to least-voted.
[...]

Issue #2732, opened 2024-08-23
[Feature Request] Add simulated vote results


CrowbarToolsRobot:
### Description
When using "Simulate Event" to simulate a "Channel Poll Ending", I should be able to add poll result names and vote counts for testing `$pollWinningChoiceName` and `$pollWinningChoiceVotes`. Currently I can only add "Title" which also doesn't seem accessible via $vars.
### Submitted By
spex.dev

> Created via Discord.

phroggster:
The `$pollWinningChoiceName` and `$pollWinningChoiceVotes` variables will gain simulated event data in an upcoming release.

It looks like we still have some work to do on getting the poll choices and the poll title exposed to variables, and I'll see what I can do about that this weekend.

phroggster:
The PR proposed adds variables for accessing the poll choices and the poll title. However, the simulate event interface is incompatible with injecting an array of poll choices, but manually using "Test Effects" will include three simulated poll options. This does **not** solve the initial request (allow manually specifying polling data in a simulated event), but it would fix the remainder of secondary complaints listed here.

phroggster:
Both of the issues brought up here should be resolved in the next release of Firebot. `$pollTitle` will allow access to the title of the poll inside of a polling event (Poll Started, Poll Update, Poll Ended). See https://github.com/crowbartools/Firebot/issues/2348#issuecomment-2427468232 for further information about the `$pollChoices` variable.

PR #2777, opened 2024-09-03
feat (vars): Add $pollTitle and $pollChoices Variables

phroggster:
### Description of the Change

- Add a `$pollTitle` variable that provides the name of a poll being started, updated, and/or ended.
- Add a `$pollChoices` variable that can return *any* piece of information about poll choices for the same event types without requiring `$objectWalkPath` or `$arrayElement`-style walking:.
- Raw `$pollChoices` will yield an array of the choices available (with `id`, `index`, `points`, `title`, `votes` properties), but accepts three optional parameters:

| Param Name | Param Description | Acceptable Values |
|---|---|---|
| `propertySelector` | Chooses which value to select from the poll choices. | count (number of choices), raw (return whole objects), index, points, title, votes |
| `sortOrder` | Changes the sorting order to allow for index (default) and total votes sorting. | "votes" for totalVotes DESC; anything else: index ASC sort. |
| `desiredIndex` | Select which poll choice item to return using 1-based indexing. | 0/null (returns all elements), or a number between 1-5 to get choice index 0-4. |

Here's a few `$pollChoices` use cases:
| Usage | Description | Example Use Case |
|---|---|---|
| `$pollChoices[count]` | Gets the count of choices listed. | There were *4* choices available in the poll. |
| `$pollChoices[votes, votes, 1]` | Gets the number of votes cast for the most-voted item. | Option x won with *173* votes cast for it. |
| `$pollChoices[index, votes, 1]` | Gets the 1-based index of the most-voted item. | Option *2* won with x votes cast for it. |
| `$pollChoices[votes, index, 1]` | Gets the number of votes cast for the first item. | Option 1 received *18* votes cast for it. |
| `$pollChoices[points]` | Gets an array of point votes used on all items. | There were `$math[$arrayJoin[$pollChoices[points], " + "]]` total channel point votes cast on all options. |

### Complaints
- Simulate event doesn't work with any of this, manual execution does. 🤷 
- 1-based array indexing is the devil, but does present a better UX.

### Applicable Issues

#2348 (partial), #2732 (partial)

### Testing

- Tested just about everything I could, but I welcome a bigger hammer to smash this with.
- However, I'm not affiliated, so a lot of this is out of my testing reach.

### Screenshots

N/A


dennisrijsdijk:
would you mind adding "partially resolves #2348" into your applicable issues?

dennisrijsdijk:
image
Noticed something when simulating event. The Choices header will show up empty due to the way the modal parses the manualMetadata.

dennisrijsdijk:
Some questions that came up as I looked into this PR further:

Can you expose the channel points vote cost?
Can you expose channel points votes for the winning choice?

phroggster:
> Can you expose the channel points vote cost?

We *could* cache the poll started event data somewhere, then do maths with it, but we'd have to do so in the twurple poll started callback, not in our start poll effect. Otherwise, the information would be missing if the poll was externally started.

> Can you expose channel points votes for the winning choice?

I've corrected my description: channel points votes is what the "points" property selector will give out. Twitch's event docs for this are so confusing though, so I'm honestly absolutely clueless as to what any of this means at a high level. Aren't all poll votes locked behind channel points now, and totalVotes essentially locked to channelPointsVotes in the event data?

dennisrijsdijk:
> Aren't all poll votes locked behind channel points now, and totalVotes essentially locked to channelPointsVotes in the event data?

poll votes, I believe, are a combination between free votes, and channelPointsVotes

phroggster:
Converting to a draft, while we get a similar interface ironed out with predictions.

phroggster:
We now have simulate events functioning properly thanks to CKY. Going to leave this as a draft until we get predictions alongside with similar functionality, but I believe this is officially ready for merging.

phroggster:
I retract my last comment. The manualMetadata change to thanks simulate event broke manually running the events. Back to the drawing board....

itsjesski (approved):

PR #2785, opened 2024-09-06
feat: add more events to Effect Queue

alfw:
### Description of the Change

- Added `Effect Queue Status` as event that trigger when the status of a queue changes. 
- Added `Effect Queue Added` as a event that triggers when a new item are added to a effect queue.
- Added a filter for  `Effect Queue Status` , `Effect Queue Added` and existing `Effect Queue Cleared`.
- Added new `effectQueueStatus` ReplaceVariable that returns the `true` or `false` or `null` based of the status of the effect queue.

### Applicable Issues

None

### Testing

- Setup test queue and Chat alert messages to see if it logs correctly
- Filter tested the same way and working

### Screenshots

- Filters:
image
image
- Variables:
image
- Add new event
image

### DIFF
--- a/src/backend/effects/queues/effect-queue-runner.js
+++ b/src/backend/effects/queues/effect-queue-runner.js
@@ -142,6 +142,10 @@ class EffectQueue {
 
         logger.debug(`Added more effects to queue ${this.id}. Current length=${this._queue.length}`);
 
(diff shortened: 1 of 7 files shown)


ebiggz (approved):
Looks great, thanks for working on this!

PR #2795, opened 2024-09-09
Feat: Show expected category when command fails due to category/game filter

Oceanity:
### Description of the Change
This change simply modifies the set-in-stone "Channel category/game isn't set to the correct category/game" response with one that attempts to pull the correct name from the Twitch API to be more specific/helpful, and will fall back to the other if the API doesn't find anything

### Applicable Issues
N/A, just noticed it while working on Firebot stuff and figured it'd be quick enough to knock out 😛

### Testing
Tested failed command due to category/game filter with expected output

### Screenshots
N/A


CaveMobster (approved):
Looking good, nice addition!

Issue #2796, opened 2024-09-10
[Bug] Offline Currency "0" Payouts Prevent Other Currencies from Offline Payouts


CrowbarToolsRobot:
### Description
Any currency with a zero offline payout rate inhibits later-modified currencies from paying out while offline.
### Steps to Reproduce
1. Add a currency with a 0 offline payout.
2. Add another currency with a 1 offline payout.
3. Wait for currency timer to engage.
### Expected Behavior
Second currency should get paid out while offline.
### Firebot Version
v5.63.2
### Submitted By
_phroggie_

> Created via Discord.

PR #2797, opened 2024-09-10
fix: Offline Currency Payout Improvements

phroggster:
### Description of the Change

- Currency payouts would stop paying out while offline once any currency had a zero offline payout or was left at the defaults.
  - Changed it to keep checking additional currencies.
- As well, the model's `offline` property is an empty string if the checkbox is checked, then unchecked in the add or edit currency model (addOrEditCurrency.js/html).
  - ~~I don't want to expand the Currency type definition in currency-access.ts, but we *might* want to.~~ Edit: I didn't.
  - Fixed this in (attempt two, in) the controller's save function to undefine the value if it's null or empty.

### Applicable Issues

#2796. #2801 

### Testing

Minimal, but yes. Verified bugs existed, then verified that this fixed them.

### Screenshots

N/A


phroggster:
Ebiggz / et. al: will you let me know if you'd rather I actually overwrite the model's empty string .offline property to a `0` or `null` in addOrEditCurrency.js when it's saving the currency instead of this approach? I feel hackish about any answer to this at this point, and am unsure of what's the least-worst solution.

phroggster:
I have reverted my prior 2801 commit, and gone ahead and fixed it properly. Retested with every variation that I could come up with to break it, and this works better, and is better for the outside world.

src/backend/currency/currency-manager.ts
```diff
@@ -264,7 +264,7 @@ class CurrencyManager {
             let basePayout = currency.payout;
             if (!connectionManager.streamerIsOnline()) {
                 if (currency.offline == null || currency.offline === 0) {
-                    return;
+                    continue;
```
phroggster, line 267: This fixes 2796

src/gui/app/directives/modals/currency/addOrEditCurrency/addOrEditCurrency.js
```diff
@@ -56,6 +56,10 @@
                     return;
                 }
 
+                if (!$ctrl.currency.offline || $ctrl.currency.offline === "") {
+                    $ctrl.currency.offline = undefined;
```
CaveMobster, line 60: This is a great solution for currencies that are created or edited, but it will not fix the issue retroactively. So we'd need to find an additional solution without having to manually update all currencies. I'm currently not sure where though, possibly on load?

phroggster: I can do it on-load, no problem (in both [backend/currency/currency-access.ts](https://github.com/phroggster/Firebot/blob/602ddd0898023657ae440c3cb3d563f07e9fac01/src/backend/currency/currency-access.ts#L50), and [gui/app/services/currency.service.js](https://github.com/phroggster/Firebot/blob/602ddd0898023657ae440c3cb3d563f07e9fac01/src/gui/app/services/currency.service.js#L13)). The existing erroneous empty strings can just hang out in `currency.json` until the user edits any currency and the file gets rewritten. Commit coming soon, once it passes muster tests.

Thanks for spotting and calling out this oversight!

phroggster: I'm marking this resolved, but I'm not exceedingly happy about *any* of the decisions that I've made to correct this. Re-writing the JSON immediately after load from one location saves from patching it up in memory in the two places immediately after both loads, and should cover **all** bases. But now we've got a read-only caching layer making changes to a file.

I need a shower.

itsjesski (approved):

PR #2802, opened 2024-09-11
Tech Debt: Load User Metadata Filters

phroggster:
### Description of the Change

- metadata-key and metadata-value filters weren't being loaded.
- This appears to have been an issue since they were created back in August.
- ...and only because they were created in .TS while every other filter is written in .JS.

### Applicable Issues

None. This is **not** an issue, just a trivial implementation oversight that is barely user-visible.

Console logging showed the following:
> debug: Registered Event Filter firebot:message-text
> warn: Could not register event filter 'undefined' because no events are specified.
> warn: Could not register event filter 'undefined' because no events are specified.
> debug: Registered Event Filter firebot:new-currency-amount

### Testing

Tested event filters both positively and negatively for key and value metadata changes.

### Screenshots

Before:
undefined filters

After:
metadata


src/backend/events/filters/builtin-filter-loader.ts
```diff
@@ -1,8 +1,11 @@
-"use strict";
+import filterManager from "./filter-manager";
+import metadataKeyFilter from "./builtin/metadata-key";
+import metadataValueFilter from "./builtin/metadata-value";
 
-const filterManager = require("./filter-manager");
+export function loadFilters() {
+    filterManager.registerFilter(metadataKeyFilter);
```
CaveMobster, line 6: In terms of consistency, I'm not sure if this is the right way to add back these filters. The effects deal with a similar mix of js and ts, so possibly it's good to have a look at how those are handled and apply the same solution here (for now).

PR #2803, opened 2024-09-11
Feat: `fileName` replace variable

Oceanity:
### Description of the Change
Adds a replace variable that takes any file path, relative or absolute, and returns the name of the file with no extension.

### Applicable Issues
N/A

### Testing
Tested many edge cases and got successful results, should work with anything that `basename` in the `path` Node package can parse the name of.

### Screenshots
N/A


CaveMobster:
Nice addition! 

So just to verify: you intend to convert `path\to\file.txt` to `file.txt` with this variable. Is that correct? Because I'd say that removes the file path, not the extension of the file, which makes the description a bit confusing.

Oceanity:
> Nice addition!
> 
> So just to verify: you intend to convert `path\to\file.txt` to `file.txt` with this variable. Is that correct? Because I'd say that removes the file path, not the extension of the file, which makes the description a bit confusing.

So, the idea is to make it `path\to\file.txt` to `file`, but someone pointed out it could be simplified with the `path` package instead of the manual regex method I had set up, and I didn't bother actually making sure it was fine afterwards 😅

Thanks for catching that, will push a fix

Oceanity:
Fixed in https://github.com/crowbartools/Firebot/pull/2803/commits/c6295877285cbb470bf7c797efead524ddaac91c

image

image

CaveMobster (approved):
Looking great, thanks for adding this!

Issue #2822, opened 2024-09-19
[Feature Request] Add Quote Effect: Return Quote ID in effectOutput


dennisrijsdijk:
**Describe the solution you'd like**
Return the quote ID for the created quote as an effect output

**Additional context**
This would allow custom quote commands to return the proper quote ID

PR #2823, opened 2024-09-19
feat: quote id effect output (#2822)

dennisrijsdijk:
### Description of the Change

Adds an effect output to the "Add Quote" effect for the quote ID of the created quote. ($&quoteId)
Converts effect to typescript

### Applicable Issues

#2822

### Testing

Ensured effect runs, properly creates quote and returns the new quote's ID as an effect output ($&quoteId)


CaveMobster (approved):
Very nice addition!

Issue #2862, opened 2024-10-18
[Bug] Case Sensitive names with moderation


CrowbarToolsRobot:
### Description
!permit command is only allowing lowercase names
### Steps to Reproduce
1.
2.
3.
### Firebot Version
v5.63.2
### Submitted By
cky_

> Created via Discord.

PR #2863, opened 2024-10-18
fix: case sensitivity in !permit moderation command

dennisrijsdijk:
### Description of the Change

Update the behavior of the !permit command to support case-insensitive input and display names

### Applicable Issues

#2862

### Testing

Ensured case-insensitive usernames and displaynames work
Ensured displaynames that are different from the username work


CaveMobster (approved):
Very nice improvement, thank you!

Issue #2861, opened 2024-10-16
[Bug] Update Channel Rewards based Sort Tags not working


arblane:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Updating the Channel Reward Pause state is not peforming as expected.

### Expected Behavior

Channel Rewards with tag 'SFX' should be updated with the set value of Pause/UnPaused.

### Steps To Reproduce

1. Create or use existing Channel Reward.
2. Add a tag, 'SFX' for example.
3. Create Preset Effect (PSE) that Updates Channel Reward by Sort Tag ('SFX') and Update Paused value is set to 'Pause'.
4. Save and Test PSE

### Version

5.63.2 and firebot-v5.63.2-nightly.24.10.03

### What operating system are you using Firebot on?

Windows 11

### Relevant log output

```shell
[2024-10-16 18:23:47.0036] - info: [v5.63.2] DBG: [Oct 16th 2024 2:23 pm] Toggle CPR - Disabled
[2024-10-16 18:23:47.0040] - error: [v5.63.2] Unhandled promise rejection 
{
  message: "Cannot read properties of undefined (reading 'includes')",
  stack: "TypeError: Cannot read properties of undefined (reading 'includes')\n" +
    '    at C:\\Users\\muskr\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\effects\\builtin\\update-channel-reward.js:271:47\n' +
    '    at Array.filter (<anonymous>)\n' +
    '    at Object.onTriggerEvent (C:\\Users\\muskr\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\effects\\builtin\\update-channel-reward.js:271:14)\n' +
    '    at C:\\Users\\muskr\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\common\\effect-runner.js:89:40\n' +
    '    at new Promise (<anonymous>)\n' +
    '    at triggerEffect (C:\\Users\\muskr\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\common\\effect-runner.js:68:12)\n' +
    '    at C:\\Users\\muskr\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\common\\effect-runner.js:147:40'
}
```

### Contact Details (Optional)

arblane

[...]

Issue #2787, opened 2024-09-07
[Bug] Disable Commands Effect not working


MattyCanny:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

One of my preset effects list was getting stuck on a "Disable Commands Effect" because the effect was unable to disable the commands with a given tag. 

### Expected Behavior

The effect should disable the commands with the tag

### Steps To Reproduce

I do not know.

### Version

5.63.2

### What operating system are you using Firebot on?

Windows 11

### Relevant log output

```shell
[2024-09-07 06:24:44.0631] - info: [v5.63.2] Websocket Overlay Connection from ::1
[2024-09-07 06:25:42.0712] - error: [v5.63.2] Unhandled promise rejection 
{
  message: "Cannot read properties of undefined (reading 'includes')",
  stack: "TypeError: Cannot read properties of undefined (reading 'includes')\n" +
    '    at C:\\Users\\Rando\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\effects\\builtin\\toggle-command.js:111:56\n' +
    '    at Array.filter (<anonymous>)\n' +
    '    at Object.onTriggerEvent (C:\\Users\\Rando\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\effects\\builtin\\toggle-command.js:111:33)\n' +
    '    at C:\\Users\\Rando\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\common\\effect-runner.js:89:40\n' +
    '    at new Promise (<anonymous>)\n' +
    '    at triggerEffect (C:\\Users\\Rando\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\common\\effect-runner.js:68:12)\n' +
    '    at C:\\Users\\Rando\\AppData\\Local\\firebot\\app-5.63.2\\resources\\app.asar\\build\\backend\\common\\effect-runner.js:147:40\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)'
}
```

phroggster:
We traced this specific instance back to the sortTag. Fixed it by creating a new sortTag, adding all of the specific commands to it, then deleting the old sortTag. No clue how/where the sortTag got corrupted, no visual indication of a missing command guid nor corruption in the tags, but it did clear up after swapping it out for a new tag.

[...]

PR #2864, opened 2024-10-18
fix: toggle-by-tag throwing error if sortTags prop is missing

dennisrijsdijk:
### Description of the Change

Update Toggle-by-tag and cooldown-by-tag functionality to handle possible missing sort tags

### Applicable Issues

#2787
#2861 

### Testing

Ensured toggle-by tag and cooldown-by tag for commands, channel rewards, timers and scheduled effect lists properly ignore any items that are missing the sortTags property


CaveMobster (approved):
Looking good!

PR #2895, opened 2024-11-18
Add deb and rpm builds for distribution

pcreech:
This adds the ability to generate rpm and deb packages as part of linux distribution.  It takes advantage of other electron-installer based providers while building an electron app.

### Description of the Change

It would be nice to have a native rpm or deb package for installing to a distribution.  This is an initial attempt to add this functionality to the build system.

### Testing

The initial testing was to ensure the app was basically functional post-install on an rpm system (fedora).  I still need to verify functionality on a debian based system, but I wanted to also start getting feedback and start conversations around approach and outputs.


dennisrijsdijk:
marking blocked until build issues resolved to prevent accidental early merges.

dennisrijsdijk:
So, after doing some more testing, setting the deb build to amd64 instead of x86_64, it looks like AppArmor is not an issue on my Ubuntu 24.04 install when installed through the .deb package. The .tar.gz currently still throws the SUID helper error.

dennisrijsdijk:
I've added a contribution PR to your fork that fixes debian builds, adds some icons and consolidates and explicitly defines some settings. All info available at pcreech#1

dennisrijsdijk:
Further testing required: does this work with -beta and -nightly.DATE builds?

zunderscore:
Converted to draft until this is complete and tested.

dennisrijsdijk:
one thing I'd like to see for this is some way to skip on the deb/rpm builds in local dev environments. On my end, I end up rolling tgz builds to one of my machines quite often, and the deb/rpm packaging tools are not available on all distros.

Issue #2898, opened 2024-11-20
[Feature Request] built-in variable api endpoint


MorningStarGG:
**Describe the solution you'd like**
can we please have an endpoint that lists all variable information for all registered variables this would be useful for many cases

PR #2899, opened 2024-11-20
feat: built in variable api

CKY-:
### Description of the Change

added an api endpoint for registered variables 

### Applicable Issues

#2898

### Testing

http://localhost:7472/api/v1/variables returns an alphabetised array of all variables

### DIFF
--- a/src/server/api/v1/controllers/variableApiController.js
+++ b/src/server/api/v1/controllers/variableApiController.js
@@ -1,0 +1,23 @@
+"use strict";
+
+const variableManager = require("../../../../backend//variables/replace-variable-manager");
+
+exports.getReplaceVariables = function(req, res) {
+    const variables = variableManager.getReplaceVariables();
+
+    if (Array.isArray(variables)) {
+        variables.sort((a, b) => {
+            const handleA = a.definition.handle.toLowerCase();
+            const handleB = b.definition.handle.toLowerCase();
+            if (handleA < handleB) {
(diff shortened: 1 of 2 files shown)


src/server/api/v1/controllers/variableApiController.js
```diff
@@ -0,0 +1,23 @@
...
+            }
+            if (handleA > handleB) {
+                return 1;
+            }
+            return 0;
+        });
+    }
+
+    res.json(variables);
+};
```
ebiggz, line 23: Couple notes on this bad boy:
- `variableManager.getReplaceVariables()` always returns an array so i dont believe there is a need for an Array.isArray check
- we can utilize the built-in string method `localeCompare` for sorting here to shorten our logic
- i dont think forcing lowercase afford us anything real here
- since sort modifies the array in place, its generally good practice to spread the original array into a new one to avoid any unintentional side effects

Altogether, this could be simplified to something like:
```js
exports.getReplaceVariables = function(req, res) {
    const sortedVariables = [...manager.getReplaceVariables()]
        .sort((a, b) => a.definition.handle.localeCompare(b.definition.handle));

    res.json(sortedVariables);
};
```

ebiggz (changes requested):
Thanks for doing this! Just a suggestion for simplification and then I think this looks good to go!

ebiggz (approved):

PR #2911, opened 2024-11-28
Reveal correct trivia answer in the end

ProditorMagnus:
Currently there is no feedback which answer was correct, so it is not possible to learn from mistakes.

### Description of the Change
After user loses in trivia, response message includes index of correct answer.

### Testing

None, low risk change.


dennisrijsdijk:
It looks like this'll only post the number/index of the correct answer, is that intentional or was the intention to include the answer as text?

ProditorMagnus:
My intention was just index. I was mostly thinking about message sentence flow, thought that index would fit the message better.

In case of multiple choice questions answer text would also work, but I think 1/2 look better than true/false in sentence.

ProditorMagnus:
Maybe ```
await twitchChat.sendChatMessage(`Sorry ${chatMessage.userDisplayName ?? username}, that is incorrect. ${question.answers.length > 2 ? 'The answer was' + question.answers[question.correctIndex - 1] + '. ' : ''}Better luck next time!`, null, chatter);
```
Not sure I should still keep it as inline condition. And not sure if `question.answers[question.correctIndex - 1]` is correct syntax, that change I am not confident to do without testing.

**Comment by ProditorMagnus** [comment]
Force push was just rebasing commit from master to v5.

**Comment by dennisrijsdijk** [comment]
Couple things:

We prefer string interpolation over string concat 
```
`this is a ${variable}`
```
rather than `"this is a " + variable`

I think the condition, or the answer finding logic should definitely be broken out into a different line.

> `${question.answers.length > 2 ? 'The answer was' + question.answers[question.correctIndex - 1] + '. ' : ''`

is it intentional there's nothing here for the false condition?

Regardless, let's see what the team thinks (whether or not number, string, or a combination of the 2 should be used) and then you can go based off their feedback

**Comment by CKY-** [comment]
adding an option to this would be better then forcing everyone to show the correct answers.. you have viewers that will abuse it.

**Review by ebiggz — CHANGES_REQUESTED** [review]
Hi @ProditorMagnus- thanks for working on this!

This will be a nice addition to the trivia game. I do think we should have it be a configurable option however, so folks can retain the previous behavior if they so desire.

Here's how that could be accomplished: 
- in `trivia.js`,  add a new setting to the game definition under `settingsCategories.chatSettings.settings` (line 314 ish):
```js
displayCorrectAnswer: {
    type: "boolean",
    title: "Display Correct Answer",
    description: "Display the correct answer when the question was answered incorrectly.",
    default: false,
    sortRank: 3
}
```
- Back over in `trivia-command.js` where the `currentQuestion` object is set (around line 234), we need to include this new setting so we can access it later when the question is answered:
```js
currentQuestion = {
    // ...
    displayCorrectAnswer: triviaSettings.settings.chatSettings.displayCorrectAnswer
};
```
- Now up in the chat-message listener, we can pull out this value from the current question:
```js
    const { /*... ,*/ displayCorrectAnswer } = currentQuestion;
```
-  Lastly, we can then reference that when building the chat message. Maybe something like:
```js
await twitchChat.sendChatMessage(
    `Sorry ${chatMessage.userDisplayName ?? username}, that is incorrect.${
        displayCorrectAnswer ? ` The answer was ${question.correctIndex}.` : ""
    } Better luck next time!`,
    null,
    chatter
);
```

ProditorMagnus:
I will try to create it as option, probably next month.

dennisrijsdijk:
Superceded by #2945

dennisrijsdijk:
Hey @ProditorMagnus, I wanted to apologize for taking over this PR from you without asking beforehand. I should have checked with you before starting my own version and closing your PR.

ProditorMagnus:
I appreciate that you did it.

PR #2913, opened 2024-11-29
chore: replace variable examples

dennisrijsdijk:
### Description of the Change
Update some replace variable examples that were incorrectly changed in #2908


ebiggz (approved):
thanks for the clean up here!

Issue #2922, opened 2024-12-12
[Feature Request] Custom currency command sends {user} can we not


CrowbarToolsRobot:
### Description
I have a command set up that requires you to have at least 20 of my custom currency. If you don't have that, it sends a custom message that says "Sorry @{user}, you don't have enough..." However, when that is actually posted to chat, it posts as "@{user} Sorry @{user}, you don't have enough..." Is there any way to get rid of that initial user callout at the beginning? It's not in my message, it seems to just be a fixed part of the response. None of the other command responses seem to do that, so it's very odd.
### Submitted By
cky_

> Created via Discord from a [message](https://discord.com/channels/372817064034959370/1316755438930497609/1316755438930497609).

CKY-:
turn off reply to or remove the name

PR #2923, opened 2024-12-12
feat: add send as reply as an option in restrictions

CKY-:
### Description of the Change

add send as reply to restrictions when not met 

### Applicable Issues

#2922

### Testing

{A1AE6F8F-BFDF-4CF4-B90E-411BF398F997}

### Screenshots

{B5C81942-CB42-495D-B8C0-3D6453E5905E}


src/types/restrictions.d.ts
```diff
@@ -1,3 +1,5 @@
+import bodyParser from "body-parser";
+
```
zunderscore, line 2: Hey, mind removing this import? Looks like it might have gotten added by mistake.

zunderscore (changes requested):
Just need to remove that one import, but otherwise like good to me

zunderscore (approved):

Issue #2398, opened 2024-02-13
[Feature Request] specify backup location 


CKY-:
**Describe the solution you'd like**
please allow a place for custom backup locations so we don't have to use a symbolic links

it would be nice to do... F:\FirebotBackup

florian-luce:
I was going to make the same request, If it can complete the request:

**Describe the solution you'd like**
In the FireBot settings menu, in the 'Backups' section, Have an option to be able to modify the default behavior for the backups folder.
If this folder no longer exists, the default value is restored.

**Additional context**
In my case, I want to outsource the backups.
I have a NAS, to which I have network access, and I would like the backups to be recorded directly on it. The option would allow you to select this network space

JRawly:
I was coming here to request this.
I've just had to reset my computer. This unfortunately deleted the appdata folder where the backups were stored.

PR #2939, opened 2024-12-22
feat/chore: backup location, more settings cleanup (#2398)

zunderscore:
### Description of the Change
Adds a setting to allow users to move their Firebot backups to another location. Also automatically migrates backup-related settings from user profile `settings,json` to `global-settings.json` after launch, as backups are global operations.

Also, pruned some outdated/no longer used settings code.

### Applicable Issues
#2398

### Testing
- Backed up to existing location
- Changed backup location
- Verified old backups were copied to new location
- Verified subsequent backups saved to new location
- Verified Open Backups Folder links pointed to new folder
- Verified user-level backup settings migrated to global settings file

### Screenshots
image


dennisrijsdijk:
How does this change deal with switching OS? eg. I import my backup from a MacOS or Linux machine into a Windows machine, is there any mechanism to ensure backups don't fail by trying to use a nix filepath, and vice versa?

zunderscore:
> How does this change deal with switching OS? eg. I import my backup from a MacOS or Linux machine into a Windows machine, is there any mechanism to ensure backups don't fail by trying to use a nix filepath, and vice versa?

It doesn't. We can't account for platform changes in the backup path, just like we can't for other paths (sounds, videos, images, etc.).

zunderscore:
Just to clarify, if a user never changes their backup location, Firebot will continue to default to the backups folder in their Firebot user data folder as it did before this change.

PR #2940, opened 2024-12-23
chore-feat: Bump Twurple, add Golden Kappa Train UI and var

phroggster:
### Description of the Change

- Bump Twurple from 7.1.0 to 7.2.1.
  - Full changelog is at https://github.com/twurple/twurple/compare/68146191075...c35088735a2eeb6, but *briefly* summarized as:
    - Lots of EventSub additions (automod, channel moderate, default reward redemptions, Golden Kappa hype trains, unban requests, *sussy* user, VIP, warnings, whispers).
    - Helix / ancillary API wrappers (mostly) related to those topics.
    - Ad break timestamps fixed.
- Add Golden Kappa Train UI and variable.
  - Sadly, these emotes are not actually visible within Firebot, since it "overwrites" the default :Kappa emote for participating users, but not the actual image URL.
  - I'm not sure that it will be possible to see this unusual emote "effect" from an external client without violating Twitch API TOS. 😢 
- Add a ***completely untested*** Golden Kappa hype train filter.

### Applicable Issues

Probably?

### Testing

UI mockups, rudimentary chat message tests only: unseen side effects *may* exist.

### Screenshots

Before and after while golden comparison
The added variable
A view of the filter


dennisrijsdijk:
May be worth also adding an event filter for golden kappa train?

phroggster:
> May be worth also adding an event filter for golden kappa train?

Give this revision an eye for me, I'm largely *unfiltered*.

dennisrijsdijk:
Looks good to me, but just to be sure, can you test it for me using a simulate event? as long as the simulate event works and the regular event sends the same data as a simulate event, it should function correctly

phroggster:
> Looks good to me, but just to be sure, can you test it for me using a simulate event? as long as the simulate event works and the regular event sends the same data as a simulate event, it should function correctly

Simulation tests worked for both true and false cases.

zunderscore (approved):
Looks good, thanks!

PR #2949, opened 2024-12-28
Integration manager refactoring & Integration tokens auto refreshing

Alastor-git:
### Description of the Change
- Typescripting IntegrationManager
- Typescription AuthAPIController
- Adding ability to pass options to AuthController to be included in requests
- Adding `client_id` and `client_secret` by default to all requests for OAuth2 code flow
- Ensuring we use `AuthData` type for all methods that interact with other modules than AuthManager
- Expanding `AuthData` type so we include timestamps for creation and expiry (`expires_at` is now a timestamp so its type is naturally restored by JSonDb)
- Fixing `refreshTokenIfExpired()` so it doesn't always think the token is still valid
- Renaming `obtainmentTimestamp` to `created_at` to be consistent with norm and some providers
- Fixing some type issues in ExtraLife integration
- Renaming `oauth` to `auth` in IntegrationData so we're consistant across the IntegrationManager
- Making integrationManager getters generic so integrations can get their data returned with the proper `FirebotSettings` and `integrationDefinition` extension types. 
- Creating an abstract integrationController generic type that integrations should extend to ensure compatibility.  
- Adding a `refreshToken()` method to `IntegrationManager` to manually ask to refresh a token if autoRefresh isn't enabled. 
- Adding a `getAuth()` method that attempts to returns a valid `LinkData` object (auth token or Account ID), auto refreshing the token if enabled and necessary. This should be the preferred method of getting access token.
- Adding a `token-refreshed` event emitted by IntegrationManager each one an integration's token is refreshed. 
- Fixing `integrationCanConnect()` and `integrationIsLinked()` methods that were pulling from the wrong object. 
- Adding Integration related type definitions. 

### Applicable Issues
- Auth tokens not refreshing for integrations
- `auth` vs `oauth` properties confusing to get auth data. 
- Need to typescript IntegrationManager stuff

### Testing
- Firebot connects/disconnects/links properly, whether token is new, is refreshed, or needs to be refreshed after off time. 
- Integration connects/disconnects/links/unlinks properly whether token is new, is refreshed, needs to be refreshed after off time, or is requested to be refreshed manually. 
- `getAuth()` works with autoRefresh for token. 
- `token-refreshed` is emitted and can be received by integration
- integration can define its own aditionnal events
- integration can use the generics properly

Not tested : 
- I don't use built in integrations, so haven't been able to test it. But there shouldn't be any breaking changes. 
- I don't think changes should break existing integrations, but would be worthwhile testing a few


Samuel38:
Created related [PR in firebot-custom-scripts-types](https://github.com/crowbartools/firebot-custom-scripts-types/pull/34) with changes that reflect these.

Alastor-git:
Sorry about the force push and history rewrite. 
I had to change the details of my account and that meant having to rewrite history before it made it onto firebot's codebase. 

Should be good now. 
And discussions we had on discord should be fixed

PR #2954, opened 2025-01-04
chore: Migrate AutoMod from PubSub to EventSub

phroggster:
### Migrate AutoMod from PubSub to EventSub

- Features lost from this:
  - Flagged message part(s) differentiation.
  - Sender's name colorization.

### Applicable Issues

None

### Testing

Zero. Zip. Zilch. But hey, PubSub has **got** to go. Hey hey, ho ho. PubSub has **got** to go.

Seriously: this will compile and start, but I have literally tested none of it.

### Screenshots

N/A


zunderscore:
@phroggster Thanks for getting this rolling! I wanna hold off on this for the moment because:

1. I don't want us to lose the flagged/blocked term data
2. [This Twurple PR](https://github.com/twurple/twurple/pull/606) will give us that data once it's merged
3. We need to test this _heavily_ because it's moderation and the start of mixing EventSub into the chat system, so there's a lot of sensitivity in play here

Also, once we get the updated Twurple code, we should start using the new AutoMod statuses to match Twitch's data format (e.g. `allowed` vs. `APPROVED`).

PR #2963, opened 2025-01-11
chore: lift api to typescript

CKY-:
### Description of the Change

lift all but the auth api and the commands api to type script

### Applicable Issues

none 

### Testing

been running against postman


src/server/api/v1/controllers/effects-api-controller.ts
```diff
@@ -96,7 +106,7 @@ async function runPresetEffectList(req, res, waitForCompletion = false) {
 
     const body = req.body || {};
     const query = req.query || {};
-    let args, username;
+    let args: ParsedQs, username: string | ParsedQs | string[] | ParsedQs[];
```
Alastor-git, line 109: This looks weird to me. 
Shouldn't `username` be a string ? 
As this is, it seems to me that we're just passing along something that could very well be a dictionary. i think there should be input validation instead here, and making sure username is only ever a string or throwing a malformed request error.

CKY-: thats the type of the query

Issue #2977, opened 2025-01-24
[Feature Request] Track OBS Item Deletions


CrowbarToolsRobot:
### Description
Can we please have a way to track deleted items in all OBS effects
### Submitted By
turtle_kat

> Created via Discord.

PR #2980, opened 2025-02-03
feat(effects): add a way to see and clear unknown sources in obs toggle effects

CKY-:
### Description of the Change

adding a ui element to show and be able to remove deleted or reindexed sources for 
show/hide obs source effect 
show/hide obs filter effect
mute/unmute obs audio effect

### Applicable Issues

#2977 

### Testing

the delete button removes the selection from the effect

### Screenshots

{33CB9883-5242-434D-AB57-B262095AC1C8}
{A3C94C7F-9FDB-4B6F-B68B-6EE422F35922}
{946E9BB8-242B-444A-8F7E-D75C9254FA20}

{84A8F97C-C291-4C33-B508-D371AB5034C9}
{35E0794F-D891-47AE-A1C3-61CEF1346DAA}
{C98D92C1-927B-423E-8BB8-88BEC3BBFB1A}

{DC384955-BAEA-41B7-9D4E-4D786846088E}
{094B4FAD-E1C2-4FB6-A921-DC717285E83C}
{DD7930C3-CF2C-4B22-9E0C-636745DE2F68}


src/backend/integrations/builtin/obs/effects/toggle-obs-source-filter.ts
```diff
@@ -30,7 +30,29 @@ export const ToggleSourceFilterEffectType: EffectType<EffectProperties> =
           categories: ["common"]
       },
       optionsTemplate: `
-    <eos-container header="Filters">
+    <eos-container ng-show="orphanedSources.length > 0">
+        <div class="effect-info alert alert-warning">
+             <p><b>Warning!</b> There are {{orphanedSources.length}} orphaned source(s) referenced by this effect. 
+             Items in <b>inactive</b> OBS profiles or scene collections <i>will</i> inaccurately show these errors.
```
CaveMobster, line 36: I'm not sure I understand what you mean by this. What are you referring to when saying `these errors`?

CKY-: im not sure what a good name for that would be. we see this only in the error in the log currently and only in the error log. 
what is a good name for that "Items" "Sources" name of thing based on view type?
the intent here is to tell the user that
 Items in inactive OBS profiles or scene collections will inaccurately show these items erroneously and should be ignored until they have changed to that OBS profile or collection.

src/backend/integrations/builtin/obs/effects/toggle-obs-source-filter.ts
```diff
@@ -76,13 +98,15 @@ export const ToggleSourceFilterEffectType: EffectType<EffectProperties> =
 
           $scope.searchText = "";
 
+          $scope.orphanedSources = [];
```
CaveMobster, line 101: I'd say, make this `missingSources` instead of `orphanedSources`, to clarify that sources aren't always orphaned but are missing for Firebot. (goes for all instances in this PR).

src/backend/integrations/builtin/obs/effects/toggle-obs-source-visibility.ts
```diff
@@ -31,7 +32,30 @@ export const ToggleSourceVisibilityEffectType: Firebot.EffectType<EffectProperti
...
+            </p>
+        </div>
+</eos-container>
+<setting-container ng-show="orphanedSources.length > 0" header="Missing Sources ({{orphanedSources.length}})" collapsed="true">
+    <div ng-repeat="sceneName in orphanedSources track by $index">
+      <div class="list-item" style="display: flex;border: 2px solid #3e4045;box-shadow: none;border-radius: 8px;padding: 5px 5px;">
+        <div class="pl-5">
+          <span>Scene: {{sceneName.sceneName}},</span>
+            <span ng-if="sceneName.sourceName != null">Name: {{sceneName.sourceName}},</span>
+            <span>Id: {{sceneName.sourceId}},</span>
```
CaveMobster, line 48: I don't think we need an ID here. Most users don't utilize this kind of data unless they come to us or are looking to solve the problem themselves, in which case the logical place would be to look at the logs.

src/backend/integrations/builtin/obs/effects/toggle-obs-source-visibility.ts
```diff
@@ -31,7 +32,30 @@ export const ToggleSourceVisibilityEffectType: Firebot.EffectType<EffectProperti
...
+                Cannot find {{orphanedSources.length}} sources in this effect. Ensure the correct profile or scene collection is loaded in OBS, and OBS is running.
+            </p>
+        </div>
+</eos-container>
+<setting-container ng-show="orphanedSources.length > 0" header="Missing Sources ({{orphanedSources.length}})" collapsed="true">
+    <div ng-repeat="sceneName in orphanedSources track by $index">
+      <div class="list-item" style="display: flex;border: 2px solid #3e4045;box-shadow: none;border-radius: 8px;padding: 5px 5px;">
+        <div class="pl-5">
+          <span>Scene: {{sceneName.sceneName}},</span>
+            <span ng-if="sceneName.sourceName != null">Name: {{sceneName.sourceName}},</span>
```
CaveMobster, line 47: This will only show for newly added sources, which for existing effects makes it difficult to understand which ones are missing:

image

Maybe we can find a way to get the source name when collecting the data of the missing sources?

CaveMobster (changes requested):

src/backend/integrations/builtin/obs/effects/toggle-obs-source-visibility.ts
CKY-: this is for backwards compatibility we currently dont store the name only the Scene and ID

src/backend/integrations/builtin/obs/effects/toggle-obs-source-visibility.ts
CKY-: that is what we store in the effect we currently only store the scene and the id and action and there is no way to ask obs what a deleted item name is. each item will only show if there it is not sent to us from obs because we didn't store it previously. we cant tell what it is only that it was true false or toggle the scene and the id and not everything is stored that way.

src/backend/integrations/builtin/obs/effects/toggle-obs-source-filter.ts
CKY-: changed

src/backend/integrations/builtin/obs/effects/toggle-obs-source-visibility.ts
CKY-: {F43554AD-5F1D-4A58-A6CE-D671715FDCD0}
if i do remove the ID this is what we will end up with

src/backend/integrations/builtin/obs/effects/toggle-obs-source-visibility.ts
CKY-: sceneName.sourceName ?? "unknown" and dont hide the name field if null? but i feel we still should show the ID

CKY-: {AF0AFC12-A27B-4F08-8FF9-E0B393B02585}
but unknown tells us the same as ID not much of anything

src/backend/integrations/builtin/obs/effects/toggle-obs-source-visibility.ts
CKY-: {EE694A15-6C34-4E10-93AF-1DE6017EE583}
if you ok with this i think this is the best of both worlds new data will only show the name and old data will show the id and Unknown name

CaveMobster: Yes this looks great!

src/backend/integrations/builtin/obs/effects/toggle-obs-source-visibility.ts
CaveMobster: I think ID unless we have a name is a good way to go about it. Thanks for looking into it!

CaveMobster (approved):

PR #2983, opened 2025-02-09
feat: add a user allow list to the moderation tool

CaveMobster:
### Description of the Change
Adds an allowlist for users in addition to the url allowlist. This is technically also solved by making a custom group and adding that custom group to the exempt list, but this is mostly meant for people who want to have a very simple setup.

### Applicable Issues
n/a

### Testing
Tested if people are or aren't allowed to post urls, adding and removing names as well as removing all of them at once. 

### Screenshots
image


dennisrijsdijk:
Would it make more sense for the allowed users to use the viewer-search-modal instead, and store users by id + username? We get the userId in all chat related events so consuming it is no issue, and storing by id instead would allow more safety in the case of someone changing their username (eg. I permit account A. Account A changes its username. Account A can no longer post URLs. After Twitch releases the old name, Account B registers that name, and now gains the ability to post URLs)

CaveMobster:
> Would it make more sense for the allowed users to use the viewer-search-modal instead, and store users by id + username? We get the userId in all chat related events so consuming it is no issue, and storing by id instead would allow more safety in the case of someone changing their username (eg. I permit account A. Account A changes its username. Account A can no longer post URLs. After Twitch releases the old name, Account B registers that name, and now gains the ability to post URLs)

Ah yes, that's a good point! I'm gonna look into that.

ebiggz (approved):
Looks great! Love the return of Cave PRs too!

PR #2990, opened 2025-02-13
fix: UI Extension Overly Strict Parsing

phroggster:
### Description of the Change

- eslint/prettier automatically insert a space in between the `function` keyword and an opening paren.
  - Converting the string `function(` into `function (`
- This conflicts with the previous hard-coded string comparison for `function(` and prevented loading extension components that are otherwise "cleanly" written.
- Switched it to using a RegExp that ignores immaterial whitespace.
- Which allows for less-strict stringified component controllers.

### Applicable Issues

#2992 

### Testing

Minimal. But I *can* successfully display and dismiss or close a custom modal from a plugin, now.

### Screenshots

None.


ebiggz (approved):
Good catch here. Thanks for tracking this down and making it far more robust

Issue #2988, opened 2025-02-11
[Feature Request] Variable for `../overlay-resources` path


CrowbarToolsRobot:
### Description
I do all my work to setup Firebot on my Macbook, but I actually stream on a Windows machine and pull my updates from a git repository. 

This is working almost perfectly but for some effects, like playing a sound, it requires you to input the path to the sound and by default it uses an absolute path, which is not the same on a Mac as it is on Windows, so I'm wondering if there's a way to use a relative file path here.

I know it's possible in the Show HTML effect, but I'm not sure if it works here because when I tried `../overlay-resources` it didn't work, so maybe the working directory for this is different than when using the Show HTML effect.
### Submitted By
cky_

> Created via Discord from a [message](https://discord.com/channels/372817064034959370/1338663604685242388/1338663604685242388).

PR #3007, opened 2025-02-19
feat(variables): Overlay-Resources folder variable. (#2988)

CKY-:
### Description of the Change

add a variable to get the system path of the overlay-resources path
allow traversal of array of paths to be used in play video, play sound, show image and obs;s path effects.

### Applicable Issues

#2988

### Testing

$overlayResourcesPath/audio/ding.wav
$overlayResourcesPath[audio, ding.wav]
$overlayResourcesPath[audio/ding.wav]
slash oriantiation dont matter on windows but could be affected on other systems 
both return 

"C:\Users\cky\AppData\Roaming\Firebot\v5\overlay-resources\audio/ding.wav"

path.resolve is um dumb and allows you to go where you should not 
$overlayResourcesPath[/etc/passwd]
n

### DIFF
--- a/src/backend/variables/builtin/utility/index.ts
+++ b/src/backend/variables/builtin/utility/index.ts
@@ -12,6 +12,7 @@ import fileRead from './file-read';
 import filesInDirectory from './files-in-directory';
 import loopCount from './loop-count';
 import loopItem from './loop-item';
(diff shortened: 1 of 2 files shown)


ebiggz (approved):
Epic, thanks for this!

Issue #3006, opened 2025-02-19
[Bug] Viewer Arrived Toggle Error


CrowbarToolsRobot:
### Description
I have implemented a "viewer arrived" event that is set to toggle each event on and off (ie. "Viewer arrived 1" set to disable self and activate "Viewer arrived 2" when the second person in chat comments, the event should toggle "Viewer arrived 2" off and enable the "Viewer arrived 3"..etc)
The purpose of the event is to reward the first chatter in twitch with 500 currency, the 2nd with 250 currency, and the third 100 currency.

What I noticed after going live was that "Viewer arrived first" kept triggering, and when I looked back into the commands, I saw that the "event" within the toggle effect will not retain the selection.
### Steps to Reproduce
1. add the event "1st viewer arrived" and save
2. add a second event "2nd viewer arrived" and save
3. in "1st viewer arrived" toggle to disable self.
4. in "1st viewer arrived" toggle to enable "2nd viewer arrived"
5. in "2nd viewer arrived" toggle to disable self.
6. in "2nd viewer arrived" toggle to enable "3rd viewer arrived"
### Expected Behavior
The toggle event effect should retain selection.
### Firebot Version
v5.64
### Submitted By
bottomshelfvodka

> Created via Discord.

phroggster:
This appears to be a problem somewhere. In order for the effect to actually work, you can:
1. Add a new Toggle Event effect to a command or an event.
2. Edit the Toggle Event in any manner (just select random things).
3. Save the Toggle Event effect.
4. Save the command or event.
    - At this point in time, the serialized data for the effect stores the selected event's **name** for some reason.
6. Edit the command or event.
7. Edit the Toggle Event effect.
8. Select a different event from the list.
9. Save the effect, and save the event.
    - At this point in time, the serialized data for the effect stores the selected event's unique **ID**.
    - As such, it will magically work.

PR #3008, opened 2025-02-20
fix: Toggle Event Effect Chooser Data Typing

phroggster:
### Description of the Change

- Loading a dropdown-select element with no values available to bind to it previously locked it into string-handling mode.
- This broke the Toggle Event effect on initial creation, which expected object-handling mode, but supplied no initial binding values.
  - As such, the initially saved effect data used the name of the event, instead of the unique ID of it.
- Added an optional one-way "value-mode" attribute to dropdown-select to force it to operate in either object or string mode, or default to prior behavior when unspecified.
- Fixed two arrow-parens.
- Beyond original scope:
  - Have the effect's group name track renaming of event sets, which would cause similar problems in the effect's UI.

### Applicable Issues

#3006

### Testing

Tested Toggle Event creation and editing thoroughly, plus a handful of others. I still need to verify whether a similar issue could occur in approximately 4 other places, at which point I'll just tack them on to here sometime tonight.

### Screenshots

Newly-created effect before:
> {"id":"c99f52f9-2875-48e5-b1c2-3295988887a1","type":"firebot:toggle-event","active":true,"toggleType":"disable","selectedGroupName":"General Events","selectedEventId":"Sub"}

Newly-created effect after:
> {"id":"1fad718c-0da3-4416-aeff-d9894407f0e9","type":"firebot:toggle-event","active":true,"toggleType":"toggle","selectedGroupName":"General Events","selectedEventId":"0eaf3180-ae07-11ed-a61c-31cd9d099aeb"}

### DIFF
--- a/src/backend/effects/builtin/toggle-event.js
+++ b/src/backend/effects/builtin/toggle-event.js
@@ -23,7 +23,7 @@ const chat = {
         </eos-container>
 
         <eos-container header="Event" pad-top="true" ng-show="effect.selectedGroupName">
-            <dropdown-select options="eventOptions[effect.selectedGroupName]" selected="effect.selectedEventId"></dropdown-select>
+            <dropdown-select options="eventOptions[effect.selectedGroupName]" selected="effect.selectedEventId" value-mode="object"></dropdown-select>
         </eos-container>
 
         <eos-container header="Toggle Action" pad-top="true">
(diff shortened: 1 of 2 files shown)


phroggster:
It looks like the other locations where a similar issue could have potentially been reached all have other means of preventing it from occurring. Either length checking, hard-coded choices, or no potential way to initialize the control in an empty manner, then later show it.

*That* makes me re-question the approach I took here, but this PR does keep the UX identical while avoiding the issue. It also allows other places to be able to change the value mode in their template should they desire to, so I'm keeping it. Let me know if you'd prefer changing the UX of this one effect instead of the added attribute, or if you've got a better idea to deal with this.

ebiggz (approved):
Thanks for this fix - getting into the weeds with angularjs' archaic attribute data binding is never a fun time so I empathize.

Issue #3000, opened 2025-02-16
[Feature Request] TTS: searchable voice dropdown


CrowbarToolsRobot:
### Description
Update the dropdown in the Text To Speech effect to allow search/filtering based on user input (eg by using ui-select)
### Submitted By
dennisoti

> Created via Discord.

PR #3012, opened 2025-02-21
feat(effects): Searchable Voices in TTS Effect (#3000)

phroggster:
### Description of the Change

- Switch the Text-To-Speech effect to using a ui-select searchable element that allows for plain text-based filtering.
- Changed the effect description to include the text "TTS" for search optimization.

### Applicable Issues

#3000

### Testing

- Old effects displayed properly in edit.
- Old effects executed and ran properly.
- New effects serialized properly.
- Old effects serialized properly.
- Verified appropriate defaults.
- Verified default was searchable in the list.

### Screenshots (before PR)
Default state of a new effect (expanded):
before

---
### Screenshots (after PR)
Default state of a new effect (collapsed):
after-defaults-collapsed

Default state of a new effect (expanded):
after-defaults-expanded

Searching:
after-canada-search

Selected:
after-linda-selected

Searching for "default":
after-default-is-searchable

### DIFF
--- a/src/backend/effects/builtin/text-to-speech.ts
+++ b/src/backend/effects/builtin/text-to-speech.ts
@@ -10,7 +10,7 @@ const effect: EffectType<{
     definition: {
         id: "firebot:text-to-speech",
         name: "Text-To-Speech",
-        description: "Have Firebot read out some text.",
+        description: "Have Firebot read some text out loud with TTS.",
         icon: "fad fa-microphone-alt",
         categories: [EffectCategory.FUN],
         dependencies: []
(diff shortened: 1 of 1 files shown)


ebiggz (changes requested):
Much needed improvement - thanks for doing this!

I do have one ask:
A number of months back I created a genericized firebot wrapper component for `<ui-select>` to cut down on the boilerplate, called `<firebot-searchable-select>`. Could we update this PR to utilize that component instead? (I realize there are many places still using ui-select directly that I have yet to refactor that I assume you originally referenced, sorry about that). 

The objects in the ttsVoice array are actually already in the correct shape for <firebot-searchable-select>, so I believe you should be able to just plop in:
```html
<firebot-searchable-select
    ng-model="effect.voiceId"
    placeholder="Select or search for a voice…"
    items="ttsVoices"
 />
 ```
The component also supports an optional "description" property on items so you could additionally adjust the item for the default voice like so:
 ```js
 {
    id: "default",
    name: "Default",
    description: "The default voice set in Settings > TTS"
}
 ```

ebiggz:
I am going to merge this in as-is and refactor to `<firebot-searchable-select>` along with the other instances that need it 👍🏻

Issue #2955, opened 2025-01-05
[Feature Request] Add custom font to the chat view in firebot


CrowbarToolsRobot:
### Description
i was wondering if theres a way to add custom font to the chat view in firebot? i have a friend with dyslexia so they use a special font and was wondering if there's a way to add there own font to it
### Submitted By
cky_

> Created via Discord from a [message](https://discord.com/channels/372817064034959370/1325381675991171082/1325381675991171082).

PR #3019, opened 2025-02-25
feat(chat UI): Custom Chat Message Fonts (#2955)

phroggster:
### Description of the Change

- Add a Use Custom Font toggle to chat settings. This expands to reveal a searching ui-select font chooser for chat message text.
- This custom font is used for message text, and not URLs, alerts, usernames, or in any other areas besides message and replying-to text.
- The ui-select-choices displays the font name, an emdash, and the stylized font name in *that* font face, in the custom size when enabled.
- Added two settings:
  - `ChatCustomFontFamily` string, with the name of the font to use.
  - `ChatCustomFontFamilyEnabled` boolean, whether or not to use a custom font face.
- The 12 default fonts are included in the available list, and more can be added through the Overlay Settings.
- Without more ancillary work in font manager, there's no way to choose font-style for the default fonts, and summernote is out of place here.
  - Custom font styles (bold, italic, bold-italic, etc.) can be loaded and work, but no 'Open Sans' Bold, or the like.

### Applicable Issues

#2955

### Testing

- Deleted in-use font in-app resets back to Open Sans
- Deleted font file from appdata profile, font remains loaded and usable
  - Restarting Firebot then falls back to Open Sans
- Long font names/excessive sizes truncate appropriately in the modal
- Replies and threading all appear appropriate
- Chat alerts, pronouns, time stamps, tooltips, URLs, and user names are all unchanged
- Spaces in font file names
- A lot of other edge cases that I can't even recall to enumerate here

### Screenshots

Chat Settings Modal:
modal preview

Replying:
reply-preview

Adjustable, and Modern/Compact friendly:
changed preview

Edit: URLs are now included in the custom font:
URLS

### DIFF
--- a/src/backend/font-manager.ts
+++ b/src/backend/font-manager.ts
@@ -3,7 +3,9 @@ import path from "path";
 import logger from "./logwrapper";
 import frontendCommunicator from "./common/frontend-communicator";
 import profileManager from "./common/profile-manager";
+import { SettingsManager } from "./common/settings-manager";
 import webServer from "../server/http-server-manager";
+import { FirebotSettingsDefaults } from "../types/settings";
 
 export enum FontFormat {
     TrueType = "truetype",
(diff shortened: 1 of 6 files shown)


CaveMobster:
This looks awesome, thanks for making this! One thing I wonder about from an accessibility perspective: do we maybe want usernames and URLs etc to be transformed as well? I can imagine that if a dyslexia font is being used, it makes sense to have everything in that font. On the other hand, it might make more sense if we had something like a "dyslexia" mode that covers the entire app. What are your thoughts on that?

CaveMobster (approved):

Issue #2723, opened 2024-08-20
[Feature Request] Vip variable


CrowbarToolsRobot:
### Description
a variable that can get the list of vips.
### Submitted By
cky_

> Created via Discord.

PR #3023, opened 2025-02-26
feat(variables): vip array variable (#2723)

CKY-:
### Description of the Change

add a variable to get an array of all vips in a channel 

### Applicable Issues

#2723

### Testing

added to chat effect and it spat out the array text of all vips

### DIFF
--- a/src/backend/roles/chat-roles-manager.ts
+++ b/src/backend/roles/chat-roles-manager.ts
@@ -11,11 +11,7 @@ interface KnownBot {
     username: string;
     channels: number;
 }
-interface vipUser {
-    userId?: string;
-    userName: string;
-    userDisplayName: string;
-}
+
 interface KnownBotServiceResponse {
     bots: Array<[string, number, number]>
 }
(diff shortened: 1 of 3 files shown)


ebiggz (changes requested):
Thanks for working on this!

I have a small suggestion to shuffle around where some of the logic is happening:
The `getVips()` method in the `chat-roles-manager` should simply return its _vips array as-is, meaning it doesn't map the array. The mapping can then be moved to the vip variable itself, before it returns the data in the evaluator. This way the types and output of the `chat-roles-manager` stays clean.

```ts
// chat-roles-manager.ts
async getVips(): Promise<BasicViewer[]> {
   return this._vips;
}

// variables/builtin/twitch/vip.ts
evaluator: async () => {
    try {
        const vips = await roleManager.getVips();
        return vips.map(v => ({
            userId: v.id,
            userName: v.username,
            userDisplayName: v.displayName
        }));
    } catch {
        // Silently fail
    }
    return [];
}
```

ebiggz (approved):
Looks great!!

Issue #2805, opened 2024-09-12
[Feature Request] Allow ignoring role(s) in $randomActiveViewer


CrowbarToolsRobot:
### Description
allow ignoring viewers in custom role(s) for $randomActiveViewer
### Submitted By
dennisoti

> Created via Discord.

PR #3027, opened 2025-02-28
feat(vars): $randomActiveViewer Enhancements (#2805)

phroggster:
### Description of the Change

- Accept either a single custom role name or an array of multiple custom role names for inclusion filtering..
- Accept either a single user name or an array of multiple user names for exclusion filtering.
- Add a parameter to exclude a single custom role name or an array of custom role names.
- Add a parameter to return either or both the username and/or id.
  - `displayName` is unavailable here though, sadly.
  - Good thing there's a variable available for that. 😁
- Uses `$arrayFrom[..., ...]` for one readable example of array parameters.
  - Which also shows that `$streamer` and `$bot`  should probably always be added to the ignored users filter.
- Squashed an historic bug in `activeUserHandler.getAllActiveUsers` that was returning user ids as `number` instead of `string`.

### Applicable Issues

#2805

### Testing

Tested quite a bit for all filtering scenarios.

### Screenshots

vars menu expanded v2

### DIFF
--- a/src/backend/chat/chat-listeners/active-user-handler.js
+++ b/src/backend/chat/chat-listeners/active-user-handler.js
@@ -87,7 +87,7 @@ exports.getRandomActiveUser = (ignoreUser = "") => {
 exports.getAllActiveUsers = () => {
     return activeUsers.keys().filter(v => !isNaN(v)).map((id) => {
         return {
-            id: parseInt(id),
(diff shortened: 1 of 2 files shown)


ebiggz (approved):
Looks good - thanks for this 👍🏻

Issue #3028, opened 2025-02-28
[Bug] playSound not unloading after play


CKY-:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

`
ERROR (Renderer) Uncaught TypeError: Cannot read properties of null (reading 'removeEventListener') [url=file:///F:/fork-firebot/Firebot/build/gui/app/directives/controls/sound-player.js] [line=94] Uncaught TypeError: Cannot read properties of null (reading 'removeEventListener')
sound-player.js:108 Uncaught TypeError: Cannot read properties of null (reading 'removeEventListener')
    at Audio.soundLoadEventHandler (sound-player.js:108:39)
`

### Expected Behavior

to unload the sound file afterplay

### Version

5.64

### What operating system are you using Firebot on?

Windows 10

### Relevant log output

```shell

```

PR #3033, opened 2025-03-02
fix(effects): sound service error (#3028)

CKY-:
### Description of the Change

change load to canplay to prevent a null error 

### Applicable Issues

#3028

### Testing

opened and closed the play sound effect


CaveMobster (approved):
Great find, thanks for fixing it!

Issue #3035, opened 2025-03-04
[Bug] Currency purge


alfw:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When trying to purge a currency it dose noting.  When checking logs it tries to purge undefined.

### Expected Behavior

Not return undefined as id when trying to purge.

### Steps To Reproduce

Open Devtools  then go to currency -> select a currency -> try to purge it. Check console

### Version

Dev v5 branch - commit: 575762f

### What operating system are you using Firebot on?

Windows 11

### Relevant log output

```shell
Devtools: 
INFO Purging currency with id undefined
```

### Contact Details (Optional)

alfw

alfw:
I already have a pending local branch with the fix for this.

PR #3037, opened 2025-03-04
fix(Currency purge)

alfw:
### Description of the Change
Removed ``.id`` so it get the actual id.

### Applicable Issues
#3035

### Testing
Purged a test currency and it's no longer returning undefined and actual purge it


CaveMobster (approved):
Awesome, thank you for fixing this!

Issue #3066, opened 2025-03-28
[Bug] Effect queues locking up.


CrowbarToolsRobot:
### Description
when an effect that causes an error or fails to run in an effect queue all effect queues stop 

for instance if you have a write file effect in a queue and the file is not found the effect queue stops all further execution

this appears to be seen more with custom script effects

### Steps to Reproduce
1. create a queue 
2. add a write file 
3. add a show image 
4. add more to the queue
### Expected Behavior
the file error should stop and exit the current execution of the currently running effect list and continue running the next list
### Firebot Version
v5.63.2
### Submitted By
cky_

> Created via Discord.

PR #3067, opened 2025-03-28
fix(queue): Effect Errors Shouldn't Break Queues (#3066)

phroggster:
### Description of the Change

- Propagate effect errors by rejecting them.
- Previous behavior broke queues anytime an effect excepted inside of one without a duration specified.

### Applicable Issues

#3066 

### Testing

"yes" -#firebot-experts


CKY-:
I have tested 4 different ways that it would brake and this fixed all 4 ways that where broken as far as i can tell.

ebiggz (approved):

Issue #3086, opened 2025-04-06
[Bug] Conditional Effect Conditions Using Zero Display as [Not Set]


phroggster:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Numeric conditions in Conditional Effect preview as [Not Set].
![Image](https://github.com/user-attachments/assets/e2f24001-dc8a-4c6e-a872-4b76310beb36)

### Expected Behavior

The effect condition preview to appear as `0` instead of `[Not Set]`.

### Steps To Reproduce

1. Add a Conditional Effect somewhere.
2. Add a numeric condition (e.g. "Command Args Count") with an expected value of `0`.
3. Save the condition.

### Version

v5.64.0-beta-1

### What operating system are you using Firebot on?

Windows 10

### Relevant log output

```shell

```

PR #3087, opened 2025-04-06
fix(effects): Condition of 0 was displaying as [Not Set] (#3086)

phroggster:
### Description of the Change

- Condition values of literal zero no longer show `[Not Set]` in the overview.
- This was overlooked in #2986, but existed prior to it.

### Applicable Issues

#3086

### Testing

Tested numerous ways to potentially break it, some with absolutely atrocious conditions, and all appeared fine.

### Screenshots

args count zero after

### DIFF
--- a/src/gui/app/directives/conditional-effect/conditionDisplay.js
+++ b/src/gui/app/directives/conditional-effect/conditionDisplay.js
@@ -47,7 +47,7 @@
 
                 function getRightSideValueDisplay() {
                     return $q(async (resolve) => {
-                        if ($ctrl.condition == null || $ctrl.condition.rightSideValue == null || $ctrl.condition.rightSideValue === "") {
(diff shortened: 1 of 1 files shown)


src/gui/app/directives/conditional-effect/conditionDisplay.js
```diff
@@ -73,19 +73,19 @@
 
                 $ctrl.$onInit = function() {
                     getRightSideValueDisplay().then((value) => {
-                        $ctrl.rightSideValueDisplay = value || "[Not Set]";
+                        $ctrl.rightSideValueDisplay = value !== null && value !== undefined ? value : "[Not Set]";
```
ebiggz, line 76: This works just fine. But note that a single `value != null` check would be sufficient as it only checks null and undefined, it's not a full falsey check (that would catch 0).

ebiggz (approved):

Issue #3084, opened 2025-04-06
settingsManager plugin settings


dennisrijsdijk:
**Describe the solution you'd like**
- add savePluginSetting and deletePluginSetting to the SettingsManager
- add getPluginSetting, savePluginSetting and deletePluginSetting to the frontend settingsService

**Additional context**
Currently, SettingsManager.getPluginSetting already exists, but cannot be used as there's no way exposed to write plugin settings.

PR #3090, opened 2025-04-06
feat(plugins): add get/save/delete PluginSettings (#3084)

dennisrijsdijk:
### Description of the Change

This PR adds `savePluginSettings` and `deletePluginSettings` functions to the SettingsManager, as well as adding the aforementioned and `getPluginSettings` to the frontend settingsService

### Applicable Issues

#3084

### Testing

Ensured in a startup script that get, save and delete functions on the backend properly set the plugin setting data
Ensured malformed input (either pluginName or settingName being null, not a string or an empty string results in a thrown error)
Ensured in a startup script that get, save and delete functions on the frontend properly set the plugin setting data

### DIFF
--- a/src/backend/common/settings-manager.ts
+++ b/src/backend/common/settings-manager.ts
@@ -25,6 +25,10 @@ class SettingsManager extends EventEmitter {
             return this.getSettingPath(settingName);
         });
 
+        frontendCommunicator.on("settings:get-plugin-setting-path", (request: { pluginName: string, settingName: string }) => {
+            return this.getPluginSettingPath(request.pluginName, request.settingName);
+        });
+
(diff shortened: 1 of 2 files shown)


ebiggz:
@dennisrijsdijk Can you clarify the purpose of these changes? Is just for letting plugins more easily persist whatever data to disk instead of doing so manually? If so, I wonder if we should reconsider this as some sort of "pluginDataStorage" system that lives in a diferent file from main app settings.

dennisrijsdijk:
Essentially, I view it as a very simple data store meant for settings. Essentially, because I saw the existing getter, I thought this was already planned and just overlooked.

Making a seperate plugin KVStore would work too

dennisrijsdijk:
going to close in favor for a new PR once the plugin KVStore is ready

Issue #3151, opened 2025-05-11
[Feature Request] Add an event for automatic channel point redemptions


TheStaticMage:
**Describe the solution you'd like**
I would like Firebot to support the eventsub "channel.channel_points_automatic_reward_redemption.add V2" as event(s). This would encompass the following redemptions:

```javascript
export type EventSubChannelAutomaticRewardType =
	| 'single_message_bypass_sub_mode'
	| 'send_highlighted_message'
	| 'random_sub_emote_unlock'
	| 'chosen_sub_emote_unlock'
	| 'chosen_modified_sub_emote_unlock';
```

This issue originally covered Twitch power-ups too (e.g. gigantify an emote) but support was added separately for these with the new `channel.bits.use` event:
- https://github.com/crowbartools/Firebot/issues/3208
- https://github.com/crowbartools/Firebot/pull/3209

**Additional context**

**Original request**
~~The current channel point redemption code supports custom redeems only. However this does not trigger any events for built-in redeems like highlighting a message or unlocking an emote, or for the bit-based redeems like on-screen celebration or gigantify an emote.~~

~~I did put together some functional code for this, but I'm really not a JavaScript developer and don't really have the time or expertise to submit a production-quality PR to this project. Here's what I hacked together if it might be in any way useful: <https://github.com/TheStaticMage/Firebot/pull/1>~~

dennisrijsdijk:
I'll double check on my end if these other events do fire if the streamer uses it, however this sounds like something dev team should judge. I don't particularly like the concept of handling the same event in two places (multi-month gifted subs is a great example of that breaking with time or when something is changed) but if dev team thinks this is a solution, I'll amend the bits.use event to ensure it ignores events for the streamer (Incase Twitch has a change of heart and starts sending out the event there too) and we can implement for the streamer here.

PR #3211, opened 2025-07-30
feat: Twitch Automatic Channel Point Redeems (#3151)

TheStaticMage:
### Description of the Change

This pull request adds support for the following Twitch Automatic channel point redeems:

- Send a message in sub-only mode
- Highlight my message
- Unlock a random emote
- Unlock a chosen emote
- Modify an emote

All of these events provide `$rewardCost` (Note 1) and `$rewardDescription` (Note 2)
Both message events provide `$redemptionMessage`
The three emote events provide `$unlockedEmoteUrl` and `$unlockedEmoteUrl`

Notes:

1. I chose to use the existing `$rewardCost` associated with custom redeems rather than implement a new variable for this. The underlying variable from Twurple is `channelPoints`.

2. `$rewardDescription` is available for the custom rewards but is not actually sent in the event. For the convenience of the user I just hard-coded this along with each of the events so things will be consistent with custom redeems.

### Applicable Issues

#3151 

### Testing

Tested all events on my channel. See screenshots.

### Screenshots

(Coming soon)

### DIFF
--- a/src/backend/events/builtin/twitch-event-source.js
+++ b/src/backend/events/builtin/twitch-event-source.js
@@ -498,6 +498,129 @@ module.exports = {
             }
         },
         {
+            id: "channel-points-redemption-single-message-bypass-sub-mode",
+            name: "Channel Points Redemption: Send a Message in Sub-Only Mode",
+            description: "When someone redeems \"Send a Message in Sub-Only Mode\" to post a message in your channel",
+            cached: false,
+            queued: false,
+            manualMetadata: {
+                username: "firebot",
+                userDisplayName: "Firebot",
+                userId: "",
+                channelPoints: 200,
+                messageText: "Test message"
+            },
+            activityFeed: {
+                icon: "fad fa-stopwatch",
+                getMessage: (eventData) => {
+                    const showUserIdName = eventData.username.toLowerCase() !== eventData.userDisplayName.toLowerCase();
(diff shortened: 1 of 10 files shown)


zunderscore:
Hey @TheStaticMage, thanks for working on this! We've had some other code changes get merged in that are causing some conflicts in your branch, so you'll need to resolve those.

Also, there seems to be some duplication/overlap with existing reward redemption logic. For example, you added a new `$redemptionMessage` which is essentially the same as the existing `$rewardMessage`. Let's use the existing variable just so we're not duplicating logic or creating confusion.

I also saw you created new logic using "points redemption" naming, but we already have "rewards redemption" files/folders that are specifically for channel point redemptive. Could you please move those things to our existing "rewards redemption" files/folders to keep them consistent?

TheStaticMage:
Re-tested and it works. Thanks for adding the markdown support by the way -- the images look much nicer now :+1: 

image

(I blurred a couple things specific to my setup that triggered with these posts so as not to distract from the content of the PR.)

zunderscore:
Hey @TheStaticMage, just FYI - did my cleanup pass. Renamed the event names to match the existing reward logic we already have (for the sake of consistency and maintenance), so once you pull latest on this or  
v5 once it gets merged, you'll need to edit your events.json to update any events you added with this. You'll need to update `channel-points-` events to `channel-reward-`.

Issue #3210, opened 2025-07-30
Startup script listing is inefficient and circular symlinks hang Firebot


TheStaticMage:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I am referring to this screen, accessed by Settings > Scripts > Manage Startup Scripts > Edit.

<img width="599" height="621" alt="Image" src="https://github.com/user-attachments/assets/b93477f5-39fc-4f92-a1ef-1fbf3207bc6b" />

There are two issues with the listing of available scripts in the dropdown:

1. It does a full recursive directory listing and _then_ filters out `node_modules` -- this can be quite time consuming on a slow system if there are a lot of under-development scripts in the tree.

2. It blindly follows symbolic links -- a circular symlink will hang Firebot.

### Expected Behavior

**I have made modifications to my local copy of Firebot to address both of these problems with a smarter recursive directory listing that does not descend into `node_modules` at all and remembers visited directories to avoid circular symlinks causing a problem. I create this issue seeking Dev Approval to submit that PR.**

### Steps To Reproduce

For slowness: When I was developing on my Android tablet under Termux it took ~10 seconds to bring up the "Edit Startup Script" modal with one project. On my PC, I have 75310 `*.js` files under my Firebot scripts directory (75190 of these are under `node_modules`) and it takes ~700 ms even with a fast system and high quality NVME drives.

For symlink: go into your scripts directory and type `ln -s . foo` and then bring up the modal. That will hang firebot.

### Version

dc0043d0e9e228ca86d26cb8fe784898b56c1694 (v5 branch 2025-07-29)

### What operating system are you using Firebot on?

Linux Distro

### Relevant log output

```shell
N/A
```

### Contact Details (Optional)

TheStaticMage on discord

PR #3240, opened 2025-09-12
fix: Improve speed and avoid hangs in custom script listing (#3210)

TheStaticMage:
### Description of the Change
This PR fixes the issues in the linked issue:
- Does not hang Firebot if there are circular symlinks in the scripts directory
- Avoids recursing into `node_modules` and `.git` directories for faster loading

### Applicable Issues
Fixes https://github.com/crowbartools/Firebot/issues/3210

### Testing
I have a super complex set of custom script layouts in my Firebot instance where there are some JS files I've imported and others that are full git repositories, and this lists them much faster than the prior code. In addition I created a circular symlink (`ln -s . foo`) in my script directory just to make sure this doesn't hang. I've been using this since I created the issue and it's worked for me on Linux. My son has also been using this on Windows.

I think this should be solid since it's a fairly straightforward use of a Set to avoid re-visiting the same path multiple times. If any of you has a Mac it may be worthwhile to verify it there too.


zunderscore (approved):
Looks awesome, thanks!

Issue #1929, opened 2022-09-12
[Feature Request] Switch statement effect


CrowbarToolsRobot:
ebiggz via Discord:

Add a switch statement effect type or perhaps a "Switch" mode to the Conditional Effect. 

Rough design:
- user would provide a value (ie `$subMonths`)
- user could create cases for this value
- user can define a "default" case.
- each case has an optional "fallthrough" option.

PR #3251, opened 2025-09-27
feat: Switch Statement Effect (#1929)

dennisrijsdijk:
### Description of the Change

This PR introduces a Switch Statement effect, which allows for simpler conditional checking with a single input value (eg. $cheerBitsAmount or $donationAmount)

In its current form, The effect compares a single input with cases that can either be string (number as string) comparisons, or number range comparisons (min <= val <= max)

The effect supports fallthrough cases, meaning if a fallthrough statement matches, it uses the effects from the next non-fallthrough case, including the default case.

The effect automatically sets a case's value or range as the label. The user can overwrite this with a custom label if desired.

### Applicable Issues

#1929

### Testing

Ensured effect loads properly and effect settings can be accessed and changed

Ensured effect shows proper error if a value is not entered

Ensured effect shows proper error if a number range is left empty or entered incorrectly (partially filled in, Min >= Max)

Ensured number range comparison properly ignores non-number values

Ensured compare mode correctly matches on identical numbers

Ensured compare mode correctly matches on identical text

Ensured compare mode correctly matches when both the case and value are empty string or null

### Screenshots

image


zunderscore (approved):
Looks great!

Issue #3285, opened 2025-10-07
Firebot hangs in an infinite loop when a preset effect list calls itself


TheStaticMage:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When I trigger a preset effect list that calls itself, it hangs Firebot.

(This was discovered by mistake, but it's possible to trivially create this situation by just having a preset effect list directly run itself.)

### Expected Behavior

Have a sanity check on the call stack depth and error out when this is exceeded, without hanging Firebot.

For example, maybe the code for a given event trigger increments a counter for a given preset effect list ID each time it gets triggered, and if this counter exceeds 10, it errors out.

### Steps To Reproduce

1. Create a preset effect list called "To infinity and beyond"
2. Add effect > Run effect list > Preset effect list > To infinity and beyond
3. Click the "play" button
4. Hang

### Version

v5 branch latest

### What operating system are you using Firebot on?

Windows 10

### Relevant log output

```shell
[2025-10-07 23:00:44.0071] - debug: Running firebot:run-effect-list(56446a47-9ff4-4e53-a32e-37240270db8d) effect...
[2025-10-07 23:00:44.0071] - debug: Looking and handling replace variables...
[2025-10-07 23:00:44.0071] - debug: Running firebot:run-effect-list(56446a47-9ff4-4e53-a32e-37240270db8d) effect...
[2025-10-07 23:00:44.0071] - debug: Looking and handling replace variables...
[2025-10-07 23:00:44.0071] - debug: Running firebot:run-effect-list(56446a47-9ff4-4e53-a32e-37240270db8d) effect...
[2025-10-07 23:00:44.0071] - debug: Looking and handling replace variables...
[2025-10-07 23:00:44.0071] - debug: Running firebot:run-effect-list(56446a47-9ff4-4e53-a32e-37240270db8d) effect...
[2025-10-07 23:00:44.0071] - debug: Looking and handling replace variables...
[2025-10-07 23:00:44.0071] - debug: Running firebot:run-effect-list(56446a47-9ff4-4e53-a32e-37240270db8d) effect...
[2025-10-07 23:00:44.0071] - debug: Looking and handling replace variables...

<forever until killed>
```

### Contact Details (Optional)

TheStaticMage

TheStaticMage:
What would you consider to be a reasonable limit @dennisrijsdijk (or anyone else)?

[...]

PR #3314, opened 2025-11-08
fix(effects): Avoid hang with self-referencing preset effect list (#3285)

TheStaticMage:
### Description of the Change
This PR eliminates Firebot hangs due to a preset effect list that directly or indirectly calls itself. After 100 calls to the same preset effect list while processing a given trigger, this logs an error and stops trying.

100 is an arbitrary "sanity check" which seemed reasonable but this can be adjusted up or down as the maintainers see fit. Doing it this way, instead of preventing any repetition, still allows reasonable recursion.

Note: There's also one linter fix my IDE made, which I left in since I've seen you cleaning up other parts of the project.

### Applicable Issues

https://github.com/crowbartools/Firebot/issues/3285

### Testing

- Preset effect list calling itself
- Preset effect list "A" that calls "B", and "B" that calls "A"

### DIFF
--- a/src/backend/effects/builtin/effect-group.js
+++ b/src/backend/effects/builtin/effect-group.js
@@ -5,6 +5,7 @@ const { EffectCategory, EffectTrigger } = require('../../../shared/effect-consta
 const { PresetEffectListManager } = require("../preset-lists/preset-effect-list-manager");
 const logger = require("../../logwrapper");
 const { simpleClone } = require("../../utils");
+const { SettingsManager } = require("../../common/settings-manager");
 
 const effectGroup = {
     definition: {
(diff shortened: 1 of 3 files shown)


dennisrijsdijk:
can we add an advanced setting to override/disable this behavior? I worry that some advanced use cases (eg. CKY's music player comes to mind) will recurse for n number of times, where n may be higher than the arbitrary safety value. I do believe for almost all users this will be a great handrail, but it'd be nice to at least still have control over it as a user.

TheStaticMage:
Would this be better as a number then, instead of a boolean? For example, 100, 1000, unlimited?

Where in the settings would you have this go? Settings > Effects or Settings > Advanced, probably?

Issue #2846, opened 2024-10-02
[Feature Request] Variables for shared chat source streamer


dennisrijsdijk:
**Describe the solution you'd like**
Add variables for the shared chat source streamer's username, displayName and id

PR #3315, opened 2025-11-12
 feat: Shared Chat Replace Variables (#2846) 

dennisrijsdijk:
### Description of the Change

This PR adds additional logic for shared chat, including a cache for the shared chat session which is updated on chat connect and when Twitch sends an EventSub event for shared chat started, updated or ended
The following variables are added:
- isSharedChatEnabled
- sharedChatParticipants
- sharedChatSourceDisplayName
- sharedChatSourceUserId
- sharedChatSourceUser
- sharedChatSourceUsername

### Applicable Issues

#2846

### Testing

Ensured Firebot starts

Ensured Firebot connects to chat

Ensured variables appear in variables menu

Ensured when starting a shared chat session, the shared chat participants are loaded into the cache and available through variables

Ensured isSharedChatEnabled and sharedChatParticipants variables are always available

Ensured sharedChatSourceUser, sharedChatSourceUsername, sharedChatSourceDisplayName and sharedChatSourceUserId are only available in chat messages and chat message events

Ensured all above variables resolve to correct data (eg. correct source channel, correct shared chat state)

Ensured when connecting to chat, the current shared chat session is fetched and loaded into the cache

Ensured the Shared Chat Session Started event fires when a shared chat session is started

Ensured the Shared Chat Session Updated event fires when a streamer joins or leaves the shared chat session

Ensured the Shared Chat Session Ended event fires when the shared chat session is ended

### DIFF
--- a/src/backend/effects/builtin/play-sound.ts
+++ b/src/backend/effects/builtin/play-sound.ts
@@ -156,6 +156,7 @@ const model: EffectType<{
         event: {
             name: "sound",
             onOverlayEvent: (event) => {
+                console.log("SOUND EVENT RECEIVED IN OVERLAY:", event);
                 const data = event;
                 const token = encodeURIComponent(data.resourceToken);
                 const resourcePath = `http://${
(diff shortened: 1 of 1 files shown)


dennisrijsdijk:
In the meantime, can I also get a review on the desired behavior for variables when a shared chat user is not found? Currently they return null. Should I do something like `[No User Found]` instead? Or maybe default to the broadcaster, considering they'd be the default source channel if there is no shared chat session.

PR #3318, opened 2025-11-13
feat: overlay alert effect

ebiggz:
A convienence effect for easily displaying alerts on the overlay (ie when someone follows). 

Effectively combines a Show Image / Play Video effect, a Show Text effect, and a Play Sound effect into one. 

Text option uses a basic text area instead of the problematic WYSIWYG editor to keep things simple. But it supports a special \<accent\> tag feature that still lets the user style subsets of the text differently.

Utilizes the new overlay position control introduced with Overlay Widget.

Intended to be a simple-to-use option that covers general use cases, with the idea being that more advanced use cases should just use the existing dedicated effects.

For now doesn't include options for youtube or twitch videos. I started work on this but the more I thought about it the less they made sense in the context of alerts so I stopped. In any case, it could be a future enhancement should there be a desire for it from users.

image

### DIFF
--- a/src/backend/chat/commands/builtin/command-management.ts
+++ b/src/backend/chat/commands/builtin/command-management.ts
@@ -1,15 +1,17 @@
 import { randomUUID } from "crypto";
 
-import { SystemCommand } from "../../../../types/commands";
-import { EffectInstance } from "../../../../types/effects";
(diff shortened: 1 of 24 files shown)


zunderscore (approved):
Hell yeah

Issue #3455, opened 2026-03-16
Viewers never individually marked offline — user:offline event not listened to


servusrene:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When a viewer leaves the chat, the `ActiveUserHandler` emits a `user:offline` event (triggered by the `_onlineUsers` NodeCache TTL expiring after 7.5 minutes of inactivity). However, no listener exists for this event, so `setChatViewerOffline()` is never called for individual viewers.

This means:
- Viewers remain `online: true` in the ViewerDB indefinitely until a bulk `setAllViewersOffline()` call occurs (app restart, chat disconnect)
- `lastSeen` keeps updating every minute for stuck-online viewers, showing today's date even if the viewer hasn't been present in days
- `minutesInChannel` (View Time) continues to accrue (+15 minutes every 15 minutes) during every live stream for viewers who are no longer watching
- The Viewer List in the UI shows ghost viewers that are not actually present

### Expected Behavior

When a viewer's online TTL expires (7.5 minutes without chat activity or chatter poll presence), they should be individually marked offline in the ViewerDB. Their `lastSeen` and `minutesInChannel` should stop updating.

### Steps To Reproduce

1. Start Firebot and go live
2. Have a viewer chat or appear in the chatter list
3. Viewer leaves Twitch entirely
4. Wait >8 minutes (past the 7.5min TTL)
5. Observe in the Viewers tab: viewer still shows today's date as "Last Seen"
6. Go live again on another day without restarting Firebot
7. Observe: viewer's View Time has increased despite not being present

### Version

All versions up to and including current master (b646e4db2)

### What operating system are you using Firebot on?

Linux Distro

### Relevant log output

```shell
No error logs — the issue is silent. The `user:offline` event fires correctly in `ActiveUserHandler` but there is simply no listener registered for it in `ViewerOnlineStatusManager`.
```

PR #3456, opened 2026-03-16
Fix: listen to user:offline event to individually mark viewers offline

servusrene:
### Description of the Change
The ActiveUserHandler emits "user:offline" when a viewer's online cache TTL expires (7.5 min without activity), but no listener existed for this event. This meant viewers were never individually marked offline in the ViewerDB — only bulk-cleared on chat disconnect or stream end.

### Applicable Issues
#3455 

### Testing
- Start Firebot, go live, have a viewer chat
- Viewer leaves Twitch
- After ~8 minutes, verify viewer is marked offline in Viewer DB
- Verify lastSeen stops updating for that viewer
- Verify view time doesn't accrue for absent viewers on next stream
- Verify bulk setAllViewersOffline() still works on chat disconnect/app restart

### DIFF
--- a/src/backend/viewers/viewer-online-status-manager.ts
+++ b/src/backend/viewers/viewer-online-status-manager.ts
@@ -41,6 +41,10 @@ class ViewerOnlineStatusManager {
         ActiveUserHandler.on("user:online", (user) => {
             void this.setChatViewerOnline(user);
         });
+
+        ActiveUserHandler.on("user:offline", (userId) => {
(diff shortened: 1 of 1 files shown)


zunderscore:
This PR is targeting the `master` branch, which should not be targeted for bugfix or feature PRs into the Firebot codebase. Please review our [contribution guidelines](https://github.com/crowbartools/Firebot/blob/master/.github%2Fcontributing.md) to ensure you are targeting the correct branch and following all policies.

PR #3463, opened 2026-03-26
fix: 3462 - Update subscriber tier on subscription event.

### LINKED ISSUE #3462: New subscribers lack tier information.
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

New subscribers are lacking information about their tier. This can cause strange behaviour when using tiers for automation, as new subscribers don't have a tier, and subscribers that upgrade their tier won't have this reflected until a reboot.

(I am just writing this cause I already have a pull request for this btw.)

### Expected Behavior

When a subscriber joins or updates their tier, their tier should also be updated in the subscriber list in the application

### Steps To Reproduce

1. Set up an automation (Like text-to-speech-polly-effect.js) based on tier 1, tier 2, tier 3 and non-subscribed users
2. With the application open, subscribe with a user that was not subscribed before.
3. Messages from the new subscriber will not be processed by the automation, as the logic will check the tier of the user before continuing, and will find nothing.

### Version

Latest

### What operating system are you using Firebot on?

Windows 10

### Relevant log output

```shell

```

### Contact Details (Optional)

_No response_

A-iko:
### Description of the Change
Updates the Twitch subscriber role cache when sub-related events occur, so new subscribers and tier changes are reflected immediately instead of only after restart. This fixes cases where subscriber-specific TTS setups would stop matching until Firebot was restarted.

### Applicable Issues
Fixes #3462

### Testing
Manual:

Reviewed the sub, resub, gift sub, gift upgrade, and Prime upgrade event paths to ensure they now update the cached subscriber tier immediately.
Confirmed the project still builds successfully with grunt prep.
Injected a subscription event to see if tiers got updated.

Regression:

Verified edited files had no diagnostics after the change.

### DIFF
--- a/src/backend/chat/chat-listeners/twitch-chat-listeners.js
+++ b/src/backend/chat/chat-listeners/twitch-chat-listeners.js
@@ -169,6 +169,13 @@ exports.setupChatListeners = (streamerChatClient, botChatClient) => {
     });
 
     streamerChatClient.onResub(async (_channel, _user, subInfo, msg) => {
+        twitchRolesManager.upsertSubscriber({
+            id: subInfo.userId,
+            username: msg.userInfo.userName,
+            displayName: msg.userInfo.displayName,
+            subTier: subInfo.plan
+        });
+
         try {
             if (subInfo.message != null && subInfo.message.length > 0) {
                 const firebotChatMessage = await chatHelpers.buildFirebotChatMessage(msg, subInfo.message);
--- a/src/backend/roles/twitch-roles-manager.ts
+++ b/src/backend/roles/twitch-roles-manager.ts
@@ -98,9 +98,49 @@ class TwitchRolesManager extends TypedEmitter<Events> {
         return this._subscribers;
     }
 
+    upsertSubscriber(subscriber: Subscriber): void {
+        if (!subscriber?.id) {
+            return;
+        }
+
+        const normalizedSubscriber: Subscriber = {
+            id: subscriber.id,
+            username: subscriber.username,
+            displayName: subscriber.displayName,
+            subTier: this.getRoleForSubTier(subscriber.subTier)
+        };
+
+        const existingSubscriber = this._subscribers.find(s => s.id === subscriber.id);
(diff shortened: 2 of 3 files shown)


zunderscore:
This PR is targeting the `master` branch, which should not be targeted for bugfix or feature PRs into the Firebot codebase. Please review our [contribution guidelines](https://github.com/crowbartools/Firebot/blob/master/.github%2Fcontributing.md) to ensure you are targeting the correct branch and following all policies.

A-iko:
> This PR is targeting the `master` branch, which should not be targeted for bugfix or feature PRs into the Firebot codebase. Please review our [contribution guidelines](https://github.com/crowbartools/Firebot/blob/master/.github%2Fcontributing.md) to ensure you are targeting the correct branch and following all policies.

Ah, I'm sorry, I must have skipped over that. I'll fix it in a bit.

servusrene:
@A-iko I've also looked into the bug (weeks before) and prepared a commit, but I haven't submitted it yet. Would you like to take a look at it?

In my opinion, the following is still missing:

- Handling of the `onChannelSubscriptionEnd` event: Remove the user from the cache when they are no longer a subscriber
- Update the subscriber list when the streamer comes online to enable long-term runtime
- Reload the list upon reconnection, even if the cache list is already full (this is necessary because events may not have been handled)

Furthermore, you could adapt the part of your code where you add a subscriber to the cache to match the mod or VIP code, making it more consistent.

EDIT: i have pushed it to gh: https://github.com/crowbartools/Firebot/pull/3465

A-iko:
@servusrene Ah, what coincidence! I'm not opposed to just letting your PR take precedence here, your changes are definitely more robust. The things you mention as missing are definitely nice additions. I'm not really very familiar with the codebase yet so my implementation was a bit minimal.

A-iko:
I'll close this PR, just going with @servusrene 's solution will save us all some time :) 
https://github.com/crowbartools/Firebot/pull/3465

Issue #3498, opened 2026-05-20
[Feature Request] Missing Months Subbed Filter for Sub (Twitch) Event


CathieNova:
**Describe the solution you'd like**
Sub (Twitch) Event is missing Months Subbed filter.
Without Months Subbed condition for filter users have to use "Conditional Effects" and this can be hard or confusing for those less experienced, and cause more clutter.

**Additional context**
No additional context as this is straight forward.

PR #3499, opened 2026-05-20
feat(filters): add months filter to Twitch sub event (#3498)

CathieNova:
### Description of the Change
Added "Months Subbed" filter to the "Sub (Twitch)" event for more flexibility with sub alerts.

The Twitch sub event already passes totalMonths and streak into the twitch:sub event, but totalMonths was not available as a filter.

The filter supports number comparisons:
"is", "is not", "less than", "less than or equal to", "greater than", and "greater than or equal to".

### Applicable Issues
[#3498](https://github.com/crowbartools/Firebot/issues/3498)

### Testing
Tested the filter both live and offline.

Confirmed:
- Firebot launches and runs normally
- The filter appears for the "Sub (Twitch)" event
- The filter settings save correctly
- The filter works as intended in a test build
- Sub events process correctly with the filter enabled

### Screenshots
image
image
image


zunderscore (approved):
This is great, thanks!

Issue #3510, opened 2026-05-24
Streak Count filter for Watch Streak event


CrowbarToolsRobot:
### Description
Add a Streak Count filter for the Watch Streak event type as an easier alternative to conditional effects
### Submitted By
ebiggz

> Created via Discord.

PR #3511, opened 2026-05-24
feat(filters): add streak count filter to watch streak event

CathieNova:
### Description of the Change
Added "Streak Count" filter to the "Watch Streak" event for more flexibility.

The Twitch Watch Streak event already passes streakCount into the twitch:watch-streak event.

The filter supports number comparisons:
"is", "is not", "less than", "less than or equal to", "greater than", and "greater than or equal to".

### Applicable Issues
[#3510](https://github.com/crowbartools/Firebot/issues/3510)

### Testing
Confirmed:
- Firebot launches and runs normally
- The filter appears for the "Watch Streak" event
- The filter settings save correctly
- The filter works as intended in a test build
- Watch Streak process correctly with the filter enabled

### Screenshots
image
image

### DIFF
--- a/src/backend/events/filters/builtin/twitch/index.ts
+++ b/src/backend/events/filters/builtin/twitch/index.ts
@@ -22,6 +22,7 @@ import subMonths from "./sub-months";
 import subType from "./sub-type";
 import treasureTrain from "./treasure-train";
 import username from "./username";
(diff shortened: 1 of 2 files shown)


ebiggz (approved):
Thanks for tackling this so fast!

Issue #3517, opened 2026-05-29
Add unit testing framework foundation


spectrum-branch:
**Describe the solution you'd like**
After discussing with the devs in the Firebot discord, it was agreed upon that it would be nice to have a unit testing framework setup in the project. The devs seemed to decide on Jest, so that is what I will be presenting in an upcoming PR.

**Additional context**
This consideration came up when the pronouns code may have undergone a regression in version 5.66.x. Having unit tests may prevent unintended regressions like this, so it would be wonderful if the project had it.

PR #3518, opened 2026-05-29
feat(testing): add jest as unit test framework (#3517)

spectrum-branch:
### Description of the Change
- Adds `jest` unit test framework to Firebot. An example unit test is staged there to allow running of tests with this foundation. This example test can be removed as "real" tests are added.
- In order to keep the linter and typescript happy, I had to create a second eslint configuration block for the test files and a separate file for the tsconfig. The production config should remain the same despite any refactoring. The reason for creating a second configuration for tests stems from `rootDir` in `tsconfig.json` being set to `'./src'`, as this does not cover the typical directory of the `./tests` folder, so npm CLI commands would complain. In no way did I want to potentially affect how production builds function, so I did not dare change `rootDir`. If the devs decide to change this value, it could lead to a bit of unification of this configuration. I researched both vitest and jest for this situation as a bit of due diligence, and they both would be affected by this.
- The jest configuration was generated using `npm init jest`, in order to create the resultant `jest.config.ts` file. This configuration can be changed as needed, and even be regenerated by running the same `npm init jest` command again and giving different answers. The below includes the whole question/answer dialogue with the generator:
```
❯ npm init jest
Need to install the following packages:
create-jest@30.4.2
Ok to proceed? (y) y

npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me
npm warn deprecated glob@10.5.0: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me

> firebotv5@5.66.5 npx
> create-jest


The following questions will help Jest to create a suitable configuration for your project

✔ Would you like to use Typescript for the configuration file? … yes
✔ Choose the test environment that will be used for testing › node
✔ Do you want Jest to add coverage reports? … no
✔ Which provider should be used to instrument code for coverage? › v8
✔ Automatically clear mock calls, instances, contexts and results before every test? … yes
```
I left behind the comments that were generated in the `jest.config.ts` file which detail the various options, but they could just as easily be removed. This is a good moment to discuss any changes to this configuration based on the preferences of the devs.

### Applicable Issues
#3517

### Testing
- Typical smoke tests should cover this, in addition to the new command `npm run test` which should run all tests. The application is intended to run normally/exactly the same as before.
- I tested by ensuring all `npm run ...` scripts worked, including `lint`, `setup`, `build`, and `start`.
```
~/git/Firebot add-unit-testing* ≡
❯ npm run lint

> firebotv5@5.66.6 lint
> grunt lint

Running "shell:eslint" (shell) task

Done.

~/git/Firebot add-unit-testing* ≡ 6s
❯ npm run test

> firebotv5@5.66.6 test
> jest

 PASS  tests/example.spec.ts
  Math tests
    ✓ should add two numbers (1 ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        0.122 s, estimated 1 s
Ran all test suites.
```

### Screenshots
image


ebiggz (approved):
Hey @spectrum-branch! Thanks for putting this together and thanks for the patience. I think this is a super solid starting point and am going to get it merged in.

Issue #3057, opened 2025-03-18
[Feature Request] Previous OBS scene variable


CrowbarToolsRobot:
### Description
Hi there, wondering if there's a way to store the previous OBS scene into a custom variable so i can create a Preset Effect List to return to whatever the previous scene was.

I can get the custom variable to change to whatever the new scene in OBS I click, but can't figure out if there's a way to store the previous scene.
### Submitted By
cky_

> Created via Discord from a [message](https://discord.com/channels/372817064034959370/1349401520067575878/1349401520067575878).

PR #3562, opened 2026-08-11
feat(vars): Add `obsPreviousSceneName` variable (#3057)

Lordmau5:
### Description of the Change
Adds a new variable called `obsPreviousSceneName` to access the previous scene that was active after switching scenes (or "" if there was no previous scene)

The value is cached similarly to the `programSceneName` in the file `obs-remote.ts`
https://github.com/crowbartools/Firebot/blob/v5/src/backend/integrations/builtin/obs/obs-remote.ts

The only thing I'm a bit unsure about is the description for the variable.

### Applicable Issues
#3057 

### Testing
This has been tested with some simple `Log Message` effects and works fine in my setup.

### Screenshots
image


ebiggz (approved):
This looks great, thanks for the contribution @Lordmau5!

PR #3576, opened 2026-09-09
Feat: horizontal chat (#3578)

dennisrijsdijk:
### Description of the Change

This PR adds a horizontal chat mode to the built-in chat widget, letting chat messages render right-to-left (default, newest message starts on the right of the widget and pushes older messages to the left) or left-to-right (reversed, newest message starts on the left of the widget and pushes older messages to the right)

### Applicable Issues
#3578

### Testing

Ensured vertical mode still behaves as it did before
Ensured messages appear horizontally and don't break into multiple lines
Ensured when using horizontal mode, the horizontal alignment option is hidden and internally forced to left for a more consistent experience with shared chat info and announcements

### Screenshots

image
image

### DIFF
--- a/src/backend/overlay-widgets/builtin-types/chat/chat.ts
+++ b/src/backend/overlay-widgets/builtin-types/chat/chat.ts
@@ -25,8 +25,7 @@ export type ChatWidgetSettings = {
     messageExitAnimation?: Animation;
     messageStyle: "compact" | "modern";
     chatOrder: "normal" | "reversed";
-    /** The type of the chat (vertical or horizontal, undefined = vertical) */
-    chatType: "vertical" | "horizontal" | undefined;
+    orientation: "vertical" | "horizontal" | undefined;
     actionDisplayFormat: "modern" | "classic";
     highlightStyle: "normal" | "highlighted";
     highlightColor?: string;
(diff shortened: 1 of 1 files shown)


src/backend/overlay-widgets/builtin-types/chat/chat.ts
```diff
@@ -25,6 +25,8 @@ export type ChatWidgetSettings = {
     messageExitAnimation?: Animation;
     messageStyle: "compact" | "modern";
     chatOrder: "normal" | "reversed";
+    /** The type of the chat (vertical or horizontal, undefined = vertical) */
+    chatType: "vertical" | "horizontal" | undefined;
```
zunderscore: Could we name this field `orientation` instead of `chatType`? I think orientation would be more precise and not need a clarifying comment.

src/backend/overlay-widgets/builtin-types/chat/chat.ts
```diff
@@ -217,17 +219,32 @@ export const chat: OverlayWidgetType<ChatWidgetSettings, ChatWidgetState> = {
...
                     label: "Reversed",
-                    description: "New messages will appear at the top of the chat feed",
+                    description: "New messages will appear at the start of the chat feed (top for vertical, left for horizontal)",
                     iconClass: "fa-sort-amount-up"
                 }
             ]
         },
+        {
+            name: "chatType",
+            title: "Chat Type",
```
zunderscore: Same as with the field. I think this would be clearer as "Orientation" rather than "Chat Type"

zunderscore (changes requested):

